PR #347, opened 2025-11-07
test: add E2E test coverage for legacy archived field format

kitelev:
## Summary
Extends E2E test coverage to verify backward compatibility with legacy `archived: true` field (in addition to `exo__Asset_isArchived: true`).

## Changes
- Added test file with `archived: true` field (legacy format)
- Added E2E test case to verify both archived field formats work correctly
- Ensures Show Archived button filters assets with either field format

## Testing
- New E2E test verifies both `archived` and `exo__Asset_isArchived` fields
- All existing tests pass

## Related
- Followup to #345
- Addresses user feedback about `archived` field support


kitelev:
Closing - will recreate PR from correct base branch

PR #2909, opened 2026-04-22
test: validate docs-only PR skips e2e (dummy — do not merge)

kitelev:
## Purpose (DO NOT MERGE)

Dummy PR to empirically validate that the path-filter gate in parent PR [#2908](https://github.com/kitelev/exocortex/pull/2908) correctly **skips** the heavy e2e jobs on a pure docs change.

**Base = `ci-speedup/phase3-path-filter`** (not main) so the filter is active in the workflow being run.

## Expected outcome

- `detect-changes`: `code=false`
- `test-component`, `e2e-shard (1..4)`, `performance-tests`: all `skipped`
- Required checks: `skipped` should satisfy branch protection (validates advisor hypothesis)
- PR mergeable status: mergeable

## Non-expected outcome (would indicate filter bug)

- Any of the gated jobs run to completion
- PR reports blocked-by-required-checks

Closing without merge once status verified.

PR #2910, opened 2026-04-22
test: R4 mitigation — docs + stray .spec.ts must trigger e2e (dummy — do not merge)

kitelev:
## Purpose (DO NOT MERGE)

R4 mitigation validation for PR [#2908](https://github.com/kitelev/exocortex/pull/2908).

Simulates the failure scenario: a docs-only PR that inadvertently includes a `.spec.ts` file (rename, migration, copy). The restrictive regex must catch this and **force** e2e to run even though the predominant change is docs.

**Base = `ci-speedup/phase3-path-filter`** so the filter is active.

## Expected outcome

- `detect-changes`: `code=true` (because `**/*.spec.{ts,tsx,js,jsx}` glob matches `docs/phase3-rename.spec.ts`)
- `test-component`, `e2e-shard (1..4)`, `performance-tests`: all run normally

## Non-expected outcome (would indicate R4 regression)

- Any of the gated jobs show `skipped` — indicates the spec-file glob in the filter is broken

Closing without merge once status verified.

### DIFF
--- a/docs/phase3-filter-validation.md
+++ b/docs/phase3-filter-validation.md
@@ -1,0 +1,10 @@
+# Phase 3 Path-Filter Validation — docs + spec rename (dummy)
+
+Dummy documentation file + a misplaced `.spec.ts` file simulating the
+R4 failure scenario (docs PR that inadvertently includes a spec
+rename/addition).
+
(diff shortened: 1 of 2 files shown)


kitelev:
Closing — same reason as #2909. Recreating against main after #2908 merges to validate R4 mitigation.

PR #3138, opened 2026-05-17
fix(shape-loader): resolve UID-form domain/range via vault UID→label map

kitelev:
## Problem (continuation of #3137)

Even after #3137 fixed `isProperty` detection for UID-form `exo__Instance_class`, SHACL still reported \`conforms: true\` on UID-canonical vaults with deliberately violating instances.

Root cause: \`wikilinkToIRI\` could not resolve pure-UID wikilinks like \`[[1b20a8f0-...]]\` (no alias, no \`<ns>__<Local>\` shape) → shape's \`domain\`/\`range\` arrays got filtered to empty → sh:class and sh:datatype constraints silently skipped during validation.

After RFC-004 + strip-canon, property frontmatter looks like:
\`\`\`yaml
exo__Property_domain: "[[1b20a8f0-d745-4e93-91db-4531b3df120e]]"   # ems__Task
exo__Property_range:  "[[1b20a8f0-d745-4e93-91db-4531b3df120e]]"   # ems__Task
\`\`\`

Validator processed 232,687 triples and emitted 0 violations even with a synthetic violator pointing the range at a wrong class.

## Fix

Two-pass scan in \`loadFromVaultFS\`:

1. **Pass 1 — \`buildUidLabelMap\`**: walks the vault, collects \`exo__Asset_uid\` → \`exo__Asset_label\` for every \`.md\` file
2. **Pass 2 — \`scanDir\`/\`processFile\`**: receives the map; \`wikilinkToIRI\` resolves pure-UID wikilinks by looking up the label and converting it to an IRI

UID detection via standard UUID v4 regex. \`uidToLabel\` is optional — when absent (e.g. \`loadFromRDFGraph\` path), existing behavior preserved.

## Test plan

- [x] New test \"resolves pure UID-form domain/range via vault uidToLabel map (post strip-canon)\" — verifies both \`domain\` AND \`range\` UID-wikilinks resolve to IRIs
- [x] All 44 ShapeLoader tests pass
- [x] Existing label-form (\`[[ems__Effort]]\`) and UID+alias form (\`[[<uid>|ems__Effort]]\`) paths unchanged

## After this merge — empirical test

I will re-run \`validate schema --shapes-mode --vault /Users/kitelev/vault-2025 --also /Users/kitelev/vault-exodev\` with the new version. If non-zero violations surface, that's empirical proof the validator is now actually checking constraints (a good thing — current zero is a false positive).

### DIFF
--- a/packages/exocortex/src/services/ShapeLoader.ts
+++ b/packages/exocortex/src/services/ShapeLoader.ts
@@ -8,6 +8,9 @@ import { Namespace } from "../domain/models/rdf/Namespace";
 const SH_NS = "http://www.w3.org/ns/shacl#";
 // XML Schema Datatypes namespace base
 const XSD_NS = "http://www.w3.org/2001/XMLSchema#";
+// UUID v4 regex (case-insensitive)
+const UUID_RE =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
 // Legacy whitelist retained for documentation; runtime resolution now goes
 // through Namespace.fromPropertyKey, which auto-extends to any well-formed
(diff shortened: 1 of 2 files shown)


kitelev:
Closing — Pass 1 (double vault scan) approach is too invasive: breaks performance-tests + cascades to e2e + test-bdd via timing. Will reimplement using loadFromRDFGraph in validate-schema.ts (reuses already-loaded tripleStore, zero extra file I/O).

Issue #3384, opened 2026-06-05
epic: Architecture cleanup program (audit 2026-06-05)


kitelev:
## Context

A 52-agent multi-agent audit with adversarial per-finding verification was run against `origin/main @ b1dd4762` on 2026-06-05. Result: **44 confirmed findings** (10 HIGH · 15 MEDIUM · 19 LOW), grouped into 4 structural themes.

Full report (local, not committed): `/Users/kitelev/Developer/exocortex-audit-2026-06-05.md`

PRs implementing items in this program should reference this epic in their description.

---

## 4 Themes

| # | Theme | LOC / scope |
|---|-------|-------------|
| A | **Dead pre-homoiconic layer** — 16 `*Button.tsx` + 10 `ICommand` + `visibility/*Rules` + 20 pinning tests. Homoiconic `exocmd__Command`/`PreconditionEvaluator` path is live; hardcoded layer is pure ballast. | ~2 400 LOC |
| B | **CLI forks core logic** in 5 places (FocusProfile resolver, AssetCreationService, folder-repair ×3, BDD grounding parser, class resolver) with confirmed active drift after production fixes. | ~1 800 LOC |
| C | **Test apparatus theater + pyramid inversion** — plugin BDD (201 scenarios) asserts its own harness state, pyramid gate can't fail on real problems, Stryker configured but wired to 0 CI workflows, visual regression `continue-on-error:true`. | — |
| D | **Docs frozen at 2026-04-13** — `docs/cli/` documents dead `exocortex command <verb>` API (0 entries in `index.ts`); `ExoQL-Specification.md` unknown to live `exo:eval`; `CLAUDE.md` lists 3 packages (reality: 5). | 4+ files |

---

### Wave 1 — Safe deletions (low risk, high LOC signal)

- [ ] **H1** — Delete pre-homoiconic cluster in one PR: 16 `*Button.tsx` (834 LOC) + 10 `ICommand` (619 LOC) + `visibility/*Rules.ts` (931 LOC) + 20 pinning tests. Pre-condition: e2e parity check per flow (start/done/vote/repair/convert/create-instance) against vault `exocmd__Command` fixtures. Risk note: grep each helper (`needsFolderRepair` etc.) before deleting — some may be live host-function callers; `CommandVisibilityContext` imported by live `MetadataExtractor`.
[...]

Issue #2810, opened 2026-04-16
fix: Create asset dropdown shows UUID filenames instead of labels (regression of #2807)


kitelev:
## Summary
`Exocortex: Create asset` (Cmd+P) dropdown renders raw UUID filenames (`7db5eeff-...md`, `1b20a8f0-...md`) instead of humanised class labels ("Task", "Project", "Area"). This is a regression of #2807 (shipped in v15.98.4) — fix patched only `AssetRelationsTable` Instance Class renderer, not the Create asset modal class-list code path.

**Re-audit reference**: F-101 in `97d88363-d103-41db-928a-d5abb36f13b2` (UI Re-audit v15.98.5 + starter-kit v1.14.0)

## Acceptance Criteria (BDD)
- [ ] **Given** plugin ≥v15.98.6 in any vault with class definitions **When** user invokes `Exocortex: Create asset` via Cmd+P **Then** dropdown shows humanised class labels ("Task", "Project", "Area", "Concept") **And** no item shows raw `<UUID>.md` filename
- [ ] **Given** the fix **When** reviewing code **Then** `ClassDiscoveryService.extractLabel()` is used consistently in both AssetRelationsTable and Create asset modal

## Technical Notes
- Root cause: Create asset modal uses independent class-list code path (likely `CreateAssetModal.ts` or `AssetCreationService`) that doesn't call `ClassDiscoveryService.extractLabel()`
- Fix #2807 patched only `AssetRelationsTable` Instance Class renderer — two surfaces, one patch
- Apply `ClassDiscoveryService.extractLabel()` to the modal's class list rendering

## Cross-references
- Regression of: #2807 (CLOSED — shipped in v15.98.4, partial fix)
- Related: #2806 (InlineTitlePatch, v15.98.5)

## Size
- [x] S (≤2h)

kitelev:
Working on fix. Root cause: NoteToRDFConverter stores exo__Asset_label as namespace IRI (via isClassReference), not literal. ClassDiscoveryService SPARQL OPTIONAL fails to bind labels. Fix: split into two mandatory-join queries + toClassName() on IRI label values.

PR #3398, opened 2026-06-05
feat(#3384): plugin Create opts into UID-canon class ref + cardinality (H3 PR2/2)

kitelev:
## Summary

**PR2/2** of epic #3384 H3. The Obsidian **Create asset** command (the button users click) now **opts into** the gated domain fields PR1 (#3397) added to core `GenericAssetCreationService`. This is a **deliberate REAL VAULT change** — the on-disk frontmatter of button-created assets changes form.

⛔ **UI-smoke gated — NOT auto-merged.** Final merge is performed by the orchestrator **after the user smoke-tests** Create-asset in real Obsidian.

### What changed
1. **`exo__Instance_class` → `[[<uuid>]]` strip-canon** (was `[[<className>]]` label form). Wired via `classRefForm: 'uuid'` + `classUid`. `classUid` is resolved during class discovery — `ClassDiscoveryService.discoverClasses()` gained a separate mandatory-join uid query (`SELECT ?class ?uid WHERE { ?class exo:Asset_uid ?uid }`, same proven split-query pattern as the #2810 label query) populating new `DiscoveredClass.classUid`. When absent → core falls back to the label form.
2. **Cardinality-aware property emission** via a SHACL-lite `shapeRegistry`, loaded **lazily on the Create command** (`ShapeLoader.loadFromRDFGraph` over the plugin's already-populated in-memory triple store — `SPARQLQueryService.getTripleStore()`). Single → scalar, Multiple → YAML array.

### ⛔ Strictly lazy — never on onload
Shapes are loaded **only** when the user runs Create (`CreateAssetCommand.loadShapeRegistry()`), guarded by `sparqlQueryService.isReady()`. The onload path is deliberately untouched (it is fragile — recent FocusProfile bootstrap hang). Shape-load failure is **non-fatal** → `undefined` → core's scalar fallback.

### Untouched (byte-identical output)
- `cli apply`, grounding `create_instance`, `ServiceRegistryPopulator`'s `createRelatedTask/Project` — none pass `classRefForm`/`shapeRegistry`. `ClassDiscoveryService` is consumed **only** by the plugin Create button.
- onload.

## Empirical verification of the shapeRegistry side-effect
When `shapeRegistry` is present, core routes **all** property values through the cardinality formatter (`String(value)` passthrough for non-wikilinks) instead of the `PropertyFieldType` formatter. I verified the on-disk impact:
- **Number / Boolean / plain-text**: **byte-identical** (`num: 5`, `flag: true` emit the same unquoted YAML either way).
- **Date-typed only**: shapeRegistry path preserves the literal value (e.g. ISO-UTC `…Z` kept; bare `2026-01-02` kept) instead of `formatTimestamp` normalization. This is **the same emission `cli create` already produces**, narrow (Date fields are rare in the create form), and arguably more faithful to user input. Flagging for reviewer/advisor awareness.

## Test plan
- [x] Behavioral golden test (`CreateAssetCommand.test.ts`) — full chain plugin-config → **real** `GenericAssetCreationService` + **real** `ShapeLoader.loadFromRDFGraph(InMemoryTripleStore)` → asserts `exo__Instance_class: "[[1b20a8f0-…]]"` (NOT `[[ems__Task]]`) + Single→scalar / Multiple→array.
- [x] **Revert-verify**: reverting the opt-in (drop `classRefForm`/`classUid`/`shapeRegistry`) → 2 tests FAIL; restore → 13/13 PASS.
- [x] `ClassDiscoveryService.test.ts` extended: `classUid` populated from uid query; undefined when no uid binding.
- [x] Other paths green unchanged: core `GenericAssetCreationService` 110/110, grounding `create_instance` green, CommandManager 60/60, typecheck 0 errors, eslint(src) clean.
- [ ] **UI-smoke in real Obsidian** (orchestrator + user) — create an asset via the button, confirm `exo__Instance_class: "[[<uuid>]]"` + cardinality, then orchestrator merges.

Epic #3384 · H3 PR2/2.

### DIFF
--- a/packages/obsidian-plugin/src/application/commands/CreateAssetCommand.ts
+++ b/packages/obsidian-plugin/src/application/commands/CreateAssetCommand.ts
@@ -3,9 +3,12 @@ import { ICommand } from "./ICommand";
 import {
   GenericAssetCreationService,
   LoggingService,
+  ShapeLoader,
   type AssetPropertyDefinition,
   type INotificationService,
+  type ShapeRegistry,
 } from "exocortex";
+import type { SPARQLQueryService } from '@plugin/application/services/SPARQLQueryService';
 import {
   showClassSelectionModal,
   type ClassSelectionModalResult,
@@ -57,6 +60,7 @@ export class CreateAssetCommand implements ICommand {
     private classDiscoveryService: ClassDiscoveryService,
     private notifier: INotificationService,
     private schemaService?: OntologySchemaService,
+    private sparqlQueryService?: SPARQLQueryService,
   ) {}
 
   /**
@@ -85,12 +89,26 @@ export class CreateAssetCommand implements ICommand {
       // Step 3: Get property definitions for formatting
       const propertyDefinitions = await this.getPropertyDefinitions(selectedClass.className);
 
-      // Step 4: Create the asset
+      // Step 3.5: Lazily load SHACL-lite shapes (#3384 H3 PR2). Built ONLY here,
+      // on the Create command — never on plugin onload (the onload path is
+      // fragile; see FocusProfile bootstrap). The plugin already has the
+      // in-memory triple store populated by the class-discovery query above, so
+      // this is a cheap read of the existing graph. Failure is non-fatal: the
+      // core service falls back to scalar (label-form / scalar) emission.
+      const shapeRegistry = await this.loadShapeRegistry();
+
+      // Step 4: Create the asset. Opt in (#3384 H3 PR2) to UID-canon class refs
+      // (`exo__Instance_class: [[<uuid>]]`) and cardinality-aware property
+      // emission. `classUid` is resolved during class discovery; when absent
+      // the core service falls back to the legacy `[[<className>]]` label form.
       const createdFile = await this.genericAssetCreationService.createAsset(
         {
           className: selectedClass.className,
+          classRefForm: "uuid",
+          classUid: selectedClass.classUid,
           label: assetResult.label || undefined,
           propertyValues: assetResult.propertyValues,
+          shapeRegistry,
         },
         propertyDefinitions,
       );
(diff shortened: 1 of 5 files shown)


kitelev:
## Round-2 advisor — caught + fixed a HIGH bug before merge

Round-1 code-reviewer: APPROVE (0 CRIT/HIGH/MEDIUM, 2 LOW). Round-2 adversarial advisor flagged a test-fixture-realism gap on the `classUid` extraction; empirical verification turned it into a **confirmed HIGH bug**, now fixed (commit `2b529638`).

### The bug
`ClassDiscoveryService` read the uid binding via `binding.get("uid")?.toString()`. In the real SPARQL engine `binding.get("uid")` is an RDF **`Literal`**, whose `.toString()` serializes to the quoted N-Triples form — empirically:
```
CTOR: Literal
toString(): "\"1b20a8f0-d745-4e93-91db-4531b3df120e\""   ← quotes are part of the string
value:      "1b20a8f0-d745-4e93-91db-4531b3df120e"        ← bare lexical form
```
So every button-created asset would have emitted a **corrupt** `exo__Instance_class: "[[\"1b20a8f0-…\"]]"`. The original unit fixtures used plain-string Map values (no-op `.toString()`), which is why both round-1 and my own tests were green. Fix: read the lexical `.value`. The fixture now uses a production-shaped `Literal`, FAILS on `.toString()` and PASSES on `.value` (revert-verified).

(The label extraction next to it is unaffected because class labels are emitted as **IRIs** — `IRI.toString()` is already bare — which is also why `className`/dropdown work in production today.)

### Corrected shapeRegistry divergence scope (round-2)
The shapeRegistry path differs from the legacy `PropertyFieldType` formatter for: **Date-typed values** (literal passthrough vs `formatTimestamp` normalization) **and free-text values containing `[[`** (quoted vs unquoted). Number / Boolean / Reference / Wikilink emit byte-identically (`MetadataHelpers.buildFileContent` serializes scalars via `String()` with no type-aware quoting). All of this matches what `cli create` already emits.

### UI-smoke targets for the orchestrator/user (before merge)
1. Create an asset of any class → confirm `exo__Instance_class: "[[<uuid>]]"` (bare uuid, **no inner quotes**).
2. Set a property whose vault shape is **`Multiple`** cardinality → confirm it emits a YAML array.
3. Set a scalar/`Single` wikilink property → confirm scalar.

Other create paths (apply, grounding `create_instance`, onload) remain byte-identical. **Still NOT auto-merged** — orchestrator merges after the UI smoke.

Issue #3748, opened 2026-06-27
fix(core): create_instance writes unquoted YAML labels → invalid frontmatter for colon-containing labels


kitelev:
## Summary

`apply create-task` / `create_instance` writes `exo__Asset_label` (and `aliases`) **unquoted**, so any label containing `: ` (colon-space) — or other YAML-significant characters — produces an asset file with **invalid YAML frontmatter**. The created file then becomes silently unparseable by the frontmatter scanner (returns `{}`), so it is invisible to `findFileByUID` / SHACL validation / metadataCache indexing.

This is a **dogfood root-cause finding**: it is the underlying cause of issue #3701 (the create wikilink-validator couldn't read a nested asset whose YAML was malformed) and of ~16 broken WBS nodes in the Dogfood-инфра project (`exoas-exodev/exodev`).

## Repro (CLI v16.156.3, vault-tbank)

```bash
npx @kitelev/exocortex-cli apply create-task <project.md> --vault <vault> \
  --input '{"label":"ZZ probe: colon-space inside label"}' --yes
```

Produced frontmatter:

```yaml
---
exo__Asset_uid: b2333d98-...
exo__Asset_label: ZZ probe: colon-space inside label
aliases:
  - ZZ probe: colon-space inside label
exo__Instance_class:
  - "[[1b20a8f0-...|ems__Task]]"
...
---
```

`js-yaml` / `python yaml.safe_load` on this frontmatter throws:

```
mapping values are not allowed here
  in "<unicode string>", line 3, column 38:
     ... Asset_label: ZZ probe: colon-...
```

Both `exo__Asset_label:` (scalar) and the `aliases:` list item are affected.

## Root cause (candidate)

The `create_instance` write path (`packages/core/src/services/GroundingExecutor.ts` → `executeCreateInstance`, and/or the shared frontmatter serializer it uses) emits scalar property values verbatim, without YAML-quoting values that require it. A plain (unquoted) YAML scalar may not contain `: ` (colon-space); it must be quoted (`"..."` with internal `"`/`\` escaped, or `'...'`). The `label` (and its `aliases` mirror) are the most common offenders because real labels often contain colons.

## Impact

- Any asset created via the homoiconic `apply`/`create_instance` path with a colon-containing label is born with broken frontmatter → invisible to UID-by-frontmatter lookup, SHACL, and Obsidian metadataCache.
[...]

PR #3780, opened 2026-06-28
feat(#3779): CLI mutation parity — homoiconic relabel (set-label) + explicit parent (set-parent)

### LINKED ISSUE #3779: CLI: homoiconic relabel (set exo__Asset_label) + explicit-parent for link-to-parent (dogfooding mutation parity)
## Context

Dogfooding goal: **all** task/project mutations should go through `exocortex-cli apply <cmd>` (homoiconic commands), not raw file edits. Today the CLI already covers most mutations via `apply`:

- status → `mark-done`, `move-to-backlog`, `move-to-todo`, `start-effort`, `set-draft-status`, `move-to-analysis`
- zone/criticality → `set-criticality-high|medium|low`
- timestamps/planning → `set-planned-start|end`, `plan-on-today`, `plan-for-evening`, `remove-*-timestamp`
- conversion → `convert-to-project`, `convert-to-task`

Two gaps still force raw `Edit` and block full CLI-first parity (and block a planned PreToolUse enforcement hook that would block direct frontmatter edits on vault task/project assets).

## Gap 1 — no command to set/change `exo__Asset_label` (relabel)

There is **no** homoiconic command to relabel an existing asset. `fix-missing-label`, `copy-label-to-aliases`, `rename-to-uid` (filename) do not cover it. Changing a task/project label today requires raw `Edit` of `exo__Asset_label` (+ the `aliases` mirror + the body `# H1`).

**Request:** a command (e.g. cliName `set-label` / `relabel`) — a `property_set`-style grounding that sets `exo__Asset_label` from `--input '{"label":"..."}'`, and keeps the `aliases` mirror consistent. Decide whether it also rewrites the body `# H1` (it is generated from label on create) or leaves the body untouched.

## Gap 2 — `link-to-parent` cannot target an explicit parent

`apply link-to-parent <path> --dry-run` passes its precondition and "would apply" **without** any `--input`, i.e. the parent is resolved by a rule rather than specified. There is no verified way to set `ems__Effort_parent` to an **arbitrary** target UID (e.g. parenting a task to a specific project just created).

**Request:** verify `link-to-parent`'s current parent-resolution; add the ability to set `ems__Effort_parent` to an explicitly specified UID via CLI — e.g. `apply link-to-parent <path> --input '{"parent":"<uid>"}'` or a dedicated `set-parent --to <uid>`. UWI must hold (child→parent only).

## Gap 3 (note, low priority / by-design?) — no generic `set-property` for existing assets

`create --property k=v` works at creation; there is no `apply set-property k=v` for existing assets. The homoiconic model intentionally prefers named commands over a generic setter. Flag for discussion only — likely **not** implemented, but worth an explicit decision so the enforcement hook knows which properties have no CLI path (those stay Edit-allowed, fail-open).

## Acceptance (req-first SDD, RFC 0003)

- requirement(proposed) → approve → implement → **revert-verify** (RED without fix / GREEN with fix) → CI green → merge → active.
- Production-shape tests (real `apply` pipeline over a fixture vault, not stubbed groundings).
- `--dry-run` preview must reflect the real write (see dry-run-preview-not-real-output rule).
- UID-canon + co-location invariants preserved.

## Why now

Closes dogfooding parity so a PreToolUse hook can block raw `Edit`s of task/project status/zone/label/parent frontmatter and force `apply`. Until #1 and #2 land, the hook must stay fail-open on label + parent (cannot block what the CLI can't yet do).

kitelev:
Closes #3779.

## What

Closes the two dogfooding gaps that forced raw `Edit` of task/project frontmatter (and blocked a planned PreToolUse enforcement hook):

- **Gap 1 — relabel:** new homoiconic `set-label` command — `apply set-label <asset> --input '{"label":"New name"}'` sets `exo__Asset_label` and appends the new label to the `aliases` mirror (set-based dedup). Body `# H1` is **left untouched** (freeform — decision documented in the issue).
- **Gap 2 — explicit parent:** new `set-parent` command — `apply set-parent <asset> --input '{"parent":"<uid>"}'` sets `ems__Effort_parent` to `"[[<uid>]]"` (wikilink form, UWI child→parent).

## How (core)

The enabling primitive is a small, additive core change in `GroundingExecutor`:

1. **`$input.<key>` named-input substitution** in `substituteVariables` — a vault grounding can bind an inputSchema-named CLI input by name (`$input.label`, `$input.parent`) instead of the single anonymous `$input`/`$value` slot. Runs *before* the bare `$input`/`$value` substitution (more-specific-first), which still resolves to `userInput.value` (backward compatible).
2. **Fail-loud placeholder gate** in `executePropertySet` moved to *after* substitution — any unresolved `$input`/`$value` (named or anonymous) fails the write loud (no half-substituted literal persisted).
3. **YAML-safe `property_set` for string-scalar properties** — values for `STRING_SCALAR_PROPERTIES` (`exo__Asset_label` / `aliases`) are routed through the conservative quote-when-needed serializer (`serializeYamlScalar`, #3748/#3750), so a label containing `: ` stays valid, parseable YAML. Other properties (timestamps, refs) are untouched.

The commands themselves are pure vault data (homoiconic):
- `set-parent` = `property_set` (`targetProperty=ems__Effort_parent`, `targetValueRef=$input.parent` → executor wraps to `"[[…]]"`).
- `set-label` = `composite` [`property_set exo__Asset_label` (`targetValueLiteral=$input.label`), `property_append aliases` (`appendExpression=$input.label`)].

## Gap 3 (decision)

Generic `apply set-property k=v` is **not** implemented — by design. The homoiconic model prefers named commands over a generic setter. Properties with no CLI path stay Edit-allowed (fail-open) for the future enforcement hook.

## Verification — req-first SDD (RFC 0003)

- **Production-shape** integration test `apply-mutation-parity.integration.test.ts` (4 cases) exercises the real `apply` pipeline (NoteToRDFConverter → CommandResolver → GroundingExecutor) over a temp fixture vault, asserting the mutated file on disk (not `--dry-run`).
- **Revert-verified:** 4/4 RED with `GroundingExecutor` reverted to origin/main, 4/4 GREEN with the fix.
- Core unit suite for the `$input.<key>` primitive (5 cases) added to `GroundingExecutor.test.ts`.
- End-to-end smoke against the real published-shape exocmd assets: `set-label` → `exo__Asset_label: "Renamed: via CLI"` + aliases synced; `set-parent` → `ems__Effort_parent: "[[…]]"`.
- Requirements: `kitelev/exoas-exo-reqs` — 2 Active, integration-bound requirements (`f7790000-3779-4aaa-…` set-parent, `f7790000-3779-4bbb-…` set-label).
- SHACL `--shapes-mode` clean on both bumped submodules.

## Submodule bumps

- `packages/exoas-exocmd` → 6 new assets (set-label/set-parent commands + groundings). No `*.expected.json` golden fixtures and no test loads the real submodule path → no fixture coupling.
- `packages/exoas-exo-reqs` → 2 requirements.

### DIFF
--- a/packages/cli/tests/integration/apply-mutation-parity.integration.test.ts
+++ b/packages/cli/tests/integration/apply-mutation-parity.integration.test.ts
@@ -1,0 +1,242 @@
+/**
+ * Issue #3779 — CLI mutation parity: homoiconic `relabel` (set exo__Asset_label
+ * + sync aliases mirror) and explicit-parent (`set-parent`).
+ *
+ * Exercises the REAL `apply` pipeline against a temp vault (no --dry-run; we
+ * read the mutated file back — `dry-run-preview-not-real-output.md`):
+ *
+ *   apply → hydrate triple store (NoteToRDFConverter) → CommandResolver →
+ *   GroundingExecutor → property_set / property_append / composite, with the
+ *   new `$input.<key>` named-input substitution (#3779) that lets a vault
+ *   grounding bind an inputSchema-named CLI input (`--input '{"label":...}'`
+ *   / `'{"parent":...}'`).
+ *
+ * Both gaps closed here previously forced raw `Edit`:
+ *   Gap 1 — no command to relabel an asset (set exo__Asset_label + aliases).
+ *   Gap 2 — link-to-parent could not target an explicit parent UID.
+ *
+ * Production-shape: the grounding assets use UID-canon, real GroundingType
+ * catalog UIDs, and the symbolic class-label form so NoteToRDFConverter emits
+ * `rdf:type exocmd:Command`/`Grounding` without needing the TBox class defs in
+ * the fixture (same technique as apply-targetvaluequery.integration.test.ts).
+ */
+import {
+  jest,
+  describe,
+  it,
+  expect,
+  beforeEach,
+  afterEach,
+} from "@jest/globals";
+import * as fs from "fs";
+import * as path from "path";
+import * as os from "os";
+
+const { applyCommand } = await import("../../src/commands/apply.js");
+
+// GroundingType catalog UIDs (packages/core/src/domain/constants/GroundingTypeUIDs.ts)
+const GT_PROPERTY_SET = "cf3bb923-f1f1-40be-b728-782844402426";
+const GT_PROPERTY_APPEND = "572f7e69-a8a1-42f6-8113-5aa65cc4b552";
+const GT_COMPOSITE = "8f9a57db-3865-4886-92fb-c5ab7f3c3fa3";
+
+const TASK_CLASS = "1b20a8f0-d745-4e93-91db-4531b3df120e"; // ems__Task
+
+// --- set-label (relabel) command + composite grounding -----------------------
+const SET_LABEL_CMD = "f7790001-0000-0000-0000-000000000001";
+const SET_LABEL_GROUNDING = "f7790001-0000-0000-0000-000000000002";
+const SET_LABEL_STEP_LABEL = "f7790001-0000-0000-0000-000000000003";
+const SET_LABEL_STEP_ALIAS = "f7790001-0000-0000-0000-000000000004";
+
+// --- set-parent command + grounding ------------------------------------------
+const SET_PARENT_CMD = "f7790002-0000-0000-0000-000000000001";
+const SET_PARENT_GROUNDING = "f7790002-0000-0000-0000-000000000002";
+
+const fm = (lines: string[]): string => ["---", ...lines, "---", ""].join("\n");
+
+const SET_LABEL_CMD_MD = fm([
+  `exo__Asset_uid: ${SET_LABEL_CMD}`,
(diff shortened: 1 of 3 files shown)


kitelev:
**Code review (mandated for GroundingExecutor changes):** code-reviewer agent → **APPROVE**, 0 CRITICAL/HIGH.

Round-2 fixes applied for the findings:
- **MEDIUM** (false-reject/clobber of a value resolving to free text containing a `$input`/`$value` substring): substitution is now a **single regex pass** (no re-scan of inserted text → no clobber) and the fail-loud gate checks the **template** (`effectiveValue`), not the substituted output. +2 unit & +1 integration regression tests.
- **LOW**: STRING_SCALAR quoting now keys off the **normalized** (prefixed) property name.
- **LOW** (property_append has no standalone unresolved-key gate): left as a documented follow-up — safe here because the set-label composite runs the gated property_set label step first and rolls back on failure.

Revert-verified RED→GREEN; 612 core + 6 integration + guards green locally.

PR #3799, opened 2026-06-29
fix(core): workflow_transition resolves any effort class data-driven — no crash on non-Task/Project/Meeting

kitelev:
## Problem (reported by Andrey, 2026-06-29)

Clicking **"Rollback to Backlog"** on an in-Doing effort asset crashed:

```
Command failed: workflow_transition: target asset has no recognised ems__* class
in exo__Instance_class (cannot resolve workflow)
```

Generic status-transition buttons (Rollback / Start / Done) render on **any** asset
carrying `ems__Effort_status`, but `GroundingExecutor.executeWorkflowTransition`
resolved the target class through a **hardcoded 3-UID map** (`CLASS_UID_TO_LABEL` =
Task / Project / Meeting). Every other effort class — `ems__Action`, `ems__Bug`,
`ems__Idea`, `ems__OfflineMeeting`, `exo__Asset`, … — is stored in UID-canon
strip-form `"[[<uid>]]"` and is **not** in that map → `resolveAssetClassFromFrontmatter`
returned `null` → fail-loud "no recognised ems__* class".

## Fix — data-driven resolution + graceful degradation

- **`WorkflowResolver.resolveForAssetOrNull(subjectIRI, classRef)`** (new) resolves a
  workflow for **any** class ref, vault-data-driven (Homoiconicity Invariant):
  1. per-asset `ems__Effort_workflow` override (works for **any** class), then
  2. built-in Task/Project/Meeting class default (a vault `ems__Workflow` ABox with
     matching `Workflow_targetClass` overrides the hardcoded fallback when present),
  3. else `null` — the class has **no applicable status workflow**.
- **`executeWorkflowTransition`** now extracts the raw class ref (any class) and treats
  a `null` workflow as a **graceful no-op** (`ExecutionResult.notApplicable`), surfaced
  by `CommandExecutionFlow` as an **info** notice — not a "Command failed" error.
- `ems__Action` (own Create/Execute/Trash one-shot lifecycle) resolves to `null` → not
  given an Effort workflow.
- `CLASS_UID_TO_LABEL` extracted to `domain/constants/WorkflowClassUids.ts` (shared,
  re-exported from `GroundingExecutor` for back-compat) so the resolver reuses it
  without an import cycle.

**Result:** no status-transition button crashes on any effort class. Classes with a
workflow transition correctly (incl. non-built-in classes via `ems__Effort_workflow`);
classes without one degrade gracefully.

## req-first SDD

`@req:7489624f-c17f-47dd-8108-3571a5809f61` (proposed → active after merge).

**Revert-verified (unit):** reverting `executeWorkflowTransition` to the pre-fix
hardcoded 3-class gate makes the two `@req:7489624f` cases in
`GroundingExecutor.workflow_transition.test.ts` go **RED** (graceful case expected
`notApplicable`; non-built-in transition expected `success`); restored → **GREEN**.
Plus production-shape `WorkflowResolver.resolveForAssetOrNull.test.ts` (real
`InMemoryTripleStore`).

## Scope notes / follow-ups (out of scope here, Path B)

- The buttons still **render** on no-workflow classes (then no-op gracefully); hiding
  them via precondition gating is a follow-up.
- Per-class `ems__Workflow` ABox resolution for arbitrary non-built-in classes, and
  default workflows for Bug/Idea/OfflineMeeting if they should be status-managed.

## Verification

- core suite (changed dirs) green; root `check:types` 0; 3 lint guards OK;
  esbuild + mobile-loadable + console-channel OK.

### DIFF
--- a/packages/core/src/domain/constants/WorkflowClassUids.ts
+++ b/packages/core/src/domain/constants/WorkflowClassUids.ts
@@ -1,0 +1,25 @@
+import { AssetClass } from "./AssetClass";
+
+/**
+ * RFC 36347daf Phase 2 — UID → AssetClass label map for the built-in status
+ * workflows. These are the classes that ship with a default workflow
+ * (Task + Project + Meeting) so a `workflow_transition` grounding can resolve
+ * them even when their `exo__Instance_class` is stored in UID-canon strip-form
+ * (`"[[<uid>]]"`, no alias) — the canonical vault representation post-#3165.
+ *
+ * NOT an exhaustive class registry: any *other* effort-bearing class gets a
+ * workflow purely from vault data — a per-asset `ems__Effort_workflow` override
+ * or a per-class `ems__Workflow` ABox (Homoiconicity Invariant). Classes with
+ * no applicable workflow resolve to `null` (graceful no-op) rather than
+ * crashing. See {@link WorkflowResolver.resolveForAssetOrNull}.
+ *
+ * Drift guards (defense in depth):
+ *   1. `GroundingExecutor.status_uid_integrity.test.ts` — UUID-shape + uniqueness.
+ *   2. Vault `validate-wikilinks` hook blocks writes to renamed class TBox files.
+ */
+export const CLASS_UID_TO_LABEL: Readonly<Record<string, string>> =
+  Object.freeze({
+    "1b20a8f0-d745-4e93-91db-4531b3df120e": AssetClass.TASK,
+    "7db5eeff-718a-49b0-8d2b-39b084a356e3": AssetClass.PROJECT,
+    "1b0a5e34-dd7f-4ead-b43a-6c7c5a5ecaca": AssetClass.MEETING,
+  });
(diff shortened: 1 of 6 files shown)


kitelev:
### Code-reviewer (APPROVE, 0 CRITICAL/HIGH) — resolutions

- **MEDIUM-1 (multi-valued `exo__Instance_class`) — FIXED** (1a961b24): the resolver now takes the full class-ref list and scans every ref, so a multi-class asset with a workflow-bearing class not listed first still resolves it. Added a production-shape multi-class test. Revert-verified again (2 `@req` cases RED→GREEN).
- **MEDIUM-2 (bare-label non-T/P/M effort subclasses now no-op instead of getting the Task fallback) — INTENDED**: this is the deliberate Homoiconic design — a class gets a status workflow only when the vault declares one (`ems__Effort_workflow` override, or a per-class `ems__Workflow` ABox), and `ems__Action` (own one-shot lifecycle) must NOT be auto-given an Effort workflow. The old bare-label Task-fallback was accidental and only worked in legacy bare-label form; in canonical UID-canon form those classes crashed. To make any such class status-managed again, add an `ems__Effort_workflow` to its assets — no code change. (Follow-up: per-class ABox resolution for arbitrary classes; default workflows for Bug/Idea/OfflineMeeting/Initiative if they should be status-managed.)
- **LOW (dead `WorkflowResolver.resolveForAsset`) — noted**: retained for test/back-compat; removal is a follow-up.

Issue #3805, opened 2026-07-02
RDF dual-IRI gap: symbolic vs file IRI forms are unbridged in the store - SPARQL class-hierarchy walks impossible, dual-form filters everywhere (successor of vault RFC 2f37bf21)


kitelev:
## Problem

One asset lives in the RDF graph under **two disconnected identifiers**:

- **file IRI** — `obsidian://vault/<path>/<uid>.md` (subject of most triples)
- **symbolic IRI** — `https://exocortex.my/ontology/<prefix>#<LocalName>` (object emitted when the wikilink target's label parses as `prefix__LocalName`)

There is **no triple bridging the two forms**. `CommandResolver` bridges them in TypeScript (`resolveClassFileIRI` label→uid→file lookup), but SPARQL cannot — so any pure-SPARQL consumer breaks on the seam.

## Fresh empirical bites (why this needs a canonical home NOW)

1. **2026-06-27 — homoiconic precondition shipped INERT:** `$target exo:Instance_class/exo:Class_superClass* exo:Prototype` is unwalkable in the real store (`Instance_class` object = symbolic IRI; `Class_superClass` subject = file IRI → path dies after one hop; `ASK { FILTER NOT EXISTS {…} }` always true). Shipped precondition was a no-op; replaced by a fragile `STRENDS(STR(?c), "Prototype")` suffix hack. Details: `rules/sparql-iri-form-pre-verify.md` addendum 2026-06-27 (req 5579ffa1, PR #3740 fix-forward).
2. **2026-07-02 — batch-closure WBS completeness check false-flagged:** `ems__Effort_status` emits in BOTH forms across assets → a `!CONTAINS(?status, "<done-uid>")` filter missed Done-in-symbolic-form → inflated "unclosed leaf" red-flags during /pmbok-close batch (12 closures). Mitigated by dual-form matching; documented in pmbok-close SKILL.md 2026-07-02.

Every new SPARQL precondition / CQ / analytics query pays this tax (dual-form filters, suffix hacks, enum-lists) — a growing class of silent always-true/always-empty bugs.

## Prior art (what this does NOT duplicate)

- #3286 (canonical IRI form for cross-vault entities) — CLOSED/completed; addressed cross-vault path duality, NOT the symbolic↔file duality.
[...]

Issue #871, opened 2025-12-13
E2E test: Verify dynamic form fields match ontology property definitions


kitelev:
### Background

The existing E2E tests for dynamic property fields (`tests/e2e/specs/dynamic-property-fields.spec.ts`) only verify **basic UI functionality** (modal opens, buttons exist, title correct). They do NOT verify the **core feature promise**: that form fields are dynamically loaded from the RDF ontology.

### Current State

**What E2E tests currently verify:**
- ✅ Modal renders with class `.exocortex-dynamic-asset-modal`
- ✅ Title shows "Create Task"
- ✅ Create/Cancel buttons exist
- ✅ First input field is visible and focusable
- ✅ Task Size dropdown exists with options
- ✅ Open in new tab toggle exists

**What E2E tests do NOT verify:**
- ❌ Form fields are actually loaded from RDF ontology
- ❌ Number of fields matches ontology property count for \`ems__Task\` class
- ❌ Property inheritance from superclasses works (\`Task\` → \`Effort\` → \`Asset\`)
- ❌ Field types match ontology \`rdfs:range\` declarations
- ❌ Deprecated properties are hidden (if \`owl:deprecated true\`)
- ❌ Required properties are marked as required (if cardinality constraints exist)

### Desired State

E2E tests should verify:
1. **Ontology query correctness**: \`OntologySchemaService.getDirectProperties()\` returns expected properties for \`ems__Task\`
2. **Property count**: Form has exactly N fields where N = properties for \`Task\` + inherited from \`Effort\` + inherited from \`Asset\`
3. **Field type mapping**: Each property's \`rdfs:range\` correctly maps to field type (text/select/wikilink/datetime/boolean)
4. **Inheritance chain**: Properties from \`Asset\` (e.g., \`exo__Asset_label\`) and \`Effort\` (e.g., \`ems__Effort_status\`) appear in Task form
5. **Predicate handling**: Verify whether \`rdfs:domain\` or \`exo__Property_domain\` is used in actual ontology

### User Impact

**Current risk:**
- E2E passes but dynamic fields may be broken
- Ontology changes don't trigger test failures
- Users may encounter forms with missing/wrong fields

**After this fix:**
- E2E tests catch ontology-form mismatches
- Property additions/removals detected automatically
- Confidence that forms reflect actual ontology state

---

### Functional Requirements

**Given** test vault contains proper ontology files with property definitions  
**When** E2E test opens "Create Task" modal  
[...]

PR #3935, opened 2026-07-25
feat(rdf): native SPARQL class-hierarchy walks for symbolic-form instances (A2, RFC 78572fa9 Phase 0)

kitelev:
## What

RFC 78572fa9 Phase 0 = **Candidate A2 "structural symbolic-space completion"** — the safe, cheap, reversible stepping-stone of the #3805 dual-IRI root-heal (independent of the larger naming-architecture migration).

`NoteToRDFConverter` now emits `<symbolic-class> exo:Class_superClass <symbolic-parent>` (+ the #871 `rdfs:subClassOf` mirror) when a **class-def** is converted, completing the parallel **symbolic** shadow node so a pure-SPARQL transitive walk is native.

## Why (#3805 dual-IRI seam)

A class reference is stored as a **symbolic** class IRI (`.../ontology/<prefix>#<LocalName>`) when the target's label parses as `prefix__LocalName`, but the structural `exo__Class_superClass` triples existed **only** on the class file's **FILE-IRI** subject. So

```sparql
SELECT ?s WHERE { ?s exo:Instance_class ?c . ?c exo:Class_superClass* <X> }
```

returned **ZERO** for the ~99.8% of assets whose `Instance_class` object is symbolic (only the zero-length `*` seed matched). TS consumers (`CommandResolver.getClassAncestorsWithDepth`, `TripleClassHierarchy`) bridge the seam internally, but **vault-declared pure-SPARQL** preconditions / Competency Queries / analytics could not — forcing `STRENDS(STR(?c),"…")` naming hacks (req 5579ffa1).

## Design

- **Hooked at class-def conversion, not enum-reference resolution.** A LEAF class — the direct class of an instance, itself never referenced as a `Class_superClass` — still gets its own symbolic superclass edge. (The instance's `exo__Instance_class` reference does **not** trigger the enum-instance emission path — verified in source; a hand-injected-triple test would have masked this gap → the production-shape test catches it.)
- **ADDITIVE** — new triples on a symbolic subject, never a second object on a single-valued predicate → **cannot create a SHACL violation** (only widens `isSubClassOf`).
- **REVERSIBLE** — drop the class-def emission block → baseline.
- **Once per class-def** — each file is converted once by `convertVault` (structural, not store-dedup-dependent).
- **Cycle-safe** — dedup + no-self-edge guard on emission; the SPARQL property-path executor is visited-set cycle-safe for the transitive `*` walk.

## No consumer regression

- `getClassAncestorsWithDepth` matches `Class_superClass` on the **FILE-IRI** subject → disjoint from A2's symbolic-subject additions (no double-count).
- `RequiredPropertyResolver` drops symbolic subjects via `uidFrom()===null`.
- Full core suite green (**8497 passed**); prototype-precondition + grounding-type submodule suites **52/52** with `exoas-exocmd` checked out; SHACL **155/155**; typecheck 0 errors.

## Revert-verify (production-shape)

New CI-gated integration suite `packages/core/tests/integration/sparql/symbolic-superclass-walk.integration.test.ts` builds the store via the **REAL** `NoteToRDFConverter.convertVault` over a fixture in-memory vault (C1 ⊑ C2 ⊑ C3, instance A classified as the **leaf** C1) — not hand-injected triples.

**@req:0cf3f6ed-8038-44d6-bf38-a3319ef09712** — disabling the class-def symbolic emission → CQ1 walk returns nothing (**RED**, 3/4 fail); restoring → returns the instance (**GREEN**, 4/4).

Req: 0cf3f6ed-8038-44d6-bf38-a3319ef09712

## Out of scope

- De-hacking the STRENDS precondition (req 5579ffa1) to the native CQ1 form — separate follow-up.
- Facet 2 (cross-form status/value consistency) — the larger root-heal, not A2.

https://claude.ai/code/session_01Lzk1cejAQeEPsj6GtFQrRo

### DIFF
--- a/packages/core/src/services/NoteToRDFConverter.ts
+++ b/packages/core/src/services/NoteToRDFConverter.ts
@@ -401,6 +401,31 @@ export class NoteToRDFConverter {
       triples.push(new Triple(subject, exoAssetLabel, labelLiteral));
     }
 
+    // A2 (RFC 78572fa9 Phase 0, req 0cf3f6ed): when THIS note is a class-def
+    // (declares `exo__Class_superClass`), mirror its superclass edges onto its
+    // own SYMBOLIC class IRI. The class file's FILE-IRI subject already gets the
+    // Class_superClass edges from the frontmatter loop above; this completes the
+    // parallel SYMBOLIC shadow node so a pure-SPARQL transitive walk
+    // `?c exo:Class_superClass* <X>` is native for the ~99.8% of assets whose
+    // `exo__Instance_class` object is the symbolic class IRI (#3805 dual-IRI seam).
+    // Hooked at class-def conversion (not at enum-reference) so a LEAF class —
+    // the direct class of an instance, never itself referenced as a superClass —
+    // still gets its own symbolic superclass edge (the instance's Instance_class
+    // reference does NOT trigger the enum-instance path). Emitted once per
+    // class-def (each file is converted once by convertVault). Additive (new
+    // triples on a symbolic subject, never a second object on a single-valued
+    // predicate) → cannot create a SHACL violation; reversible (drop this block).
+    if (frontmatter["exo__Class_superClass"] != null) {
+      const ownSymbolicIRI =
+        this.expandClassValue(file.basename) ??
+        (typeof rawLabel === "string"
+          ? this.expandClassValue(rawLabel)
+          : null);
+      if (ownSymbolicIRI) {
+        this.emitSymbolicSuperClassEdges(ownSymbolicIRI, frontmatter);
+      }
+    }
+
     // Flush rdf:type triples emitted by valueToRDFObject for enum instance IRIs (Fix #ff3858e5)
     triples.push(...this.pendingExtraTriples);
     this.pendingExtraTriples = [];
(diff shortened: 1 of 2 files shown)


kitelev:
### Code-review (auto-merge discipline, parser path)

**Verdict: APPROVE** — 0 CRITICAL / 0 HIGH, 1 MEDIUM (consciously deferred), 2 LOW (documented-acceptable).

Reviewer independently: confirmed additive/SHACL-safety (symbolic subject is a distinct node from the file-IRI subject — verified via `store.match`; nothing single-valued gets a second object); confirmed the #871 mirror subject, `pendingExtraTriples` lifecycle (no cross-note leak), cycle-safety (converter is non-recursive), and a **genuine** revert-verify (3/4 RED when the emission is disabled, GREEN restored); ran 316 adjacent converter/hierarchy tests — all green.

**MEDIUM — deferred to #3936:** incremental live-edit re-index leaves a stale symbolic superclass edge until a full refresh (`VaultRDFIndexer.removeFileTriples` is file-IRI-subject-scoped). Deferred because: (a) A2 is Phase-0 groundwork — **no live consumer** of the pure-SPARQL hierarchy walk yet (STRENDS hack still live), so blast radius is currently nil; (b) it is a **pre-existing** bug-class (reified-statement + enum `rdf:type` emissions have the same latent behaviour) whose correct fix is **uniform**, not A2-only; (c) self-heals on any full refresh. Documented in the `emitSymbolicSuperClassEdges` docstring; the uniform fix (purge non-file-IRI emissions on incremental re-index) is tracked in #3936 to land with the first live consumer.

kitelev:
**Closing — A2's eager-materialization approach is superseded by a query-time direction (Andrey's decision).**

Empirical real-vault validation of A2 (measured with vs without the emission, on vault-my + vault-exodev via the CLI inference path):

| | A2-OFF | A2-ON |
|---|---|---|
| vault-my inferred triples | 3,663 | 38,445 |
| vault-my store size | 130,537 | 165,859 (**+27%**) |
| vault-my SHACL violations | 7 | 7 (**0 new** — pre-existing) |
| vault-exodev inferred triples | 27 | 31,138 |
| vault-exodev store size | 142,600 | 174,283 (**+22%**) |
| vault-exodev SHACL violations | 0 | 0 |

**Finding (missed by round-1 review, caught in round-2 + verified independently):** A2 is NOT the "inert additive groundwork" the requirement framed it as. Because `RDFSInferenceEngine` reads `exo__Class_superClass` on ANY subject and runs on the live plugin store (`VaultRDFIndexer.runInference` on init/refresh/prototype-edit) + the CLI, the symbolic edges A2 emits **activate global materialization of inferred ancestor `Instance_class` triples** for the ~99.8% symbolic-form instances — the #3805 root-heal effect, delivered eagerly (+22-27% store growth on every index/refresh, memory-heavy on mobile).

It is **safe** (0 new SHACL violations on both vaults) and **correct RDFS**, but its mechanism (eager materialization → store grows) is the **opposite** of the root-heal design principle Andrey set: *"store = uid only; resolve at the query boundary; keep the store minimal"* (the generalized Candidate-E query-time resolver).

**Decision:** fix Facet-1 (class-hierarchy walk) the same way as names — at query-time (generalize the E-bridge to walk the hierarchy at match-time), NOT via store materialization. That keeps the store minimal, unifies the resolution mechanism, and makes the root-heal (Candidate B) coherent rather than patchwork. Req 0cf3f6ed → Deprecated; the query-time hierarchy fix will be its own requirement when designed.

Preserved: the empirical method (build store via real `convertVault` + measure inference delta) and this finding are recorded in RFC 78572fa9 + #3936.

Issue #3520, opened 2026-06-13
feat(cli): `create` should co-locate asset by exo__Asset_isDefinedBy (currently defaults to 01 Inbox/)


kitelev:
## Summary
CLI `create` writes new assets to `01 Inbox/<uid>.md` even when `exo__Asset_isDefinedBy` resolves to an on-disk ontology folder. The co-location invariant (CLAUDE.md, RFC 0b7a2fad CR-1) is enforced by the Obsidian-plugin Write hook + `/exocortex-asset` skill + `audit co-location`, but the **CLI `create` command** does not apply the same placement resolver.

## Discovered
EKA Alpha e2e CLI validation (2026-06-14, PMBOK f33732f4 M3.3). Created E2E-TEST asset with resolvable `isDefinedBy: [[32d2374c]]` ($exodev) → landed in `01 Inbox/`, not `assetspaces/kitelev/exoas-exodev/exodev/`. Frontmatter fully correct; only physical placement wrong.

## Impact
Low / non-blocking for alpha (frontmatter correct; plugin `create` co-locates correctly; workaround = `apply repair-folder` after CLI create). But CLI-first onboarders get inbox-placement → later co-location audit churn.

## Fix
Wire `folderRepairHelpers.findReferencedFile` (packages/cli/src/executors/folderRepairHelpers.ts — same resolver as `apply repair-folder`) into CLI `create` placement. OR document that CLI `create` requires a subsequent `apply repair-folder`.

## Refs
PMBOK f33732f4 M3.3 e2e finding #1. Priority: low (post-alpha).

kitelev:
## Empirical reconfirmation (2026-06-14, Alpha Launch WBS digitization)

Reconfirmed on current published CLI: `create` **still lands assets in `01 Inbox/`** despite resolvable `exo__Asset_isDefinedBy` (dry-run `--class 1b20a8f0 --property exo__Asset_isDefinedBy=[[32d2374c]]` → `path: 01 Inbox/<uid>.md`). The documented workaround (`apply repair-folder` after create) is **insufficient for submodule-co-located ontologies**: target folder `assetspaces/kitelev/exoas-exodev/exodev/` is a git **submodule**, so the repair-move is a cross-submodule boundary move (root `01 Inbox/` → submodule) — gitlink-risky if automated (`git mv` breaks gitlinks; see rule submodule-cross-boundary-move).

[...]

Issue #3488, opened 2026-06-12
bug(shacl): false-positive violations from IRI-representation artifacts (4 mechanisms)


kitelev:
## Summary

SHACL-lite validator produces false-positive violations from IRI-representation artifacts rather than real data problems. Empirically verified during vault-2025 SHACL-cleanup session (2026-06-12): after exhaustive data cleanup (281 → 99 violations; 182 removed were genuine data issues), ALL remaining 99 violations are engine artifacts — not data. Single umbrella issue covering 4 co-rooted mechanisms.

## User Story / JTBD

**As a** vault maintainer running `validate schema --shapes-mode`
**I want** reported violations to reflect only genuine data problems
**So that** I can trust the validator output and clean data confidently without chasing phantom issues

## Background

**Related (all CLOSED — informational links, not blockers):**
- #3281 `feat(sparql): cross-vault index + runtime materialization` (CLOSED)
- #3284 `feat(validate): cross-vault SHACL constraint checking via --also flag` (CLOSED)
- #3286 `refactor(sparql): canonical IRI form for cross-vault entities` (CLOSED)

The Exocortex IRI representation has two parallel schemes that co-exist post-UUID-canon (RFC-004 2026-05-16):
- **Symbolic** (when target label is parseable as `prefix__LocalName`): `<https://exocortex.my/ontology/<prefix>#<LocalName>>`
- **Filename-form**: `<obsidian://vault/<path>/<uuid>.md>`

Which form a wikilink emits depends on the *target's* `exo__Asset_label` parseability, and whether the wikilink includes an alias (`[[<uid>|prefix__Local]]` → symbolic; `[[<uid>]]` → filename). This dual-scheme is correct by design — but the SHACL validator does not bridge these forms consistently, producing 4 distinct false-positive patterns totaling 99 violations on a fully clean dataset.

## Related Issues

- **Related (informational):** #3281, #3284, #3286 (all CLOSED — predecessors that partially addressed cross-vault and IRI issues)

## BoK References

| Body of Knowledge | Chapter/Section | Relevance |
|-------------------|-----------------|-----------|
| SWEBOK v3 | Ch. 4 Software Testing | Validator correctness: distinguishing false positives from true violations |
| SWEBOK v3 | Ch. 5 Software Maintenance | Bug fix: validator engine mis-classification of valid data |
| DMBOK v2 | Ch. 13 Data Quality | False-positive violation = wrong data quality signal |
[...]

Issue #3245, opened 2026-05-23
validate schema --staged ignores scope filter, reports all vault violations


kitelev:
## Summary

`validate schema --shapes-mode --staged` is documented as "Only validate git-staged .md files (for pre-commit hooks)", but in practice it ignores the staged scope filter and reports **all** vault violations regardless of git state. With an empty staged set, exit code is still `1` and full vault violation list is emitted.

This blocks the use of SHACL-lite validation in pre-commit hooks — every commit would fail on accumulated unrelated violations even when the changed files are clean.

## Repro

```bash
cd /Users/kitelev/vault-2025
git reset HEAD                     # empty staged set
git diff --cached --name-only | wc -l   # → 0
npx @kitelev/exocortex-cli@16.20.2 validate schema --shapes-mode --staged --format json | jq '.data.violationCount'
# → 283
echo "exit=$?"
# → exit=1
```

## Expected

- With 0 staged files: `violationCount = 0`, `conforms = true`, `exit 0`
- With N staged files: only violations whose `focusNode` (or related source asset) is in the staged set

## Actual

- Returns full vault violation count (283 in our case), `conforms = false`, `exit 1`

## Environment

- Version: `@kitelev/exocortex-cli@16.20.2` (also reproduced in 16.19.3)
- Vault: ~12k .md files, ~283 baseline violations
- Platform: macOS 25.4.0, node 16.19.3

## Impact

Currently the project has `scripts/validate-task-context.js` doing narrow `sh:class` validation for a single hardcoded property (`ems__Task_context`). Switching pre-commit hook to call `validate schema --shapes-mode --staged` would deliver generic coverage of all property ranges — but it's blocked by this scope-filter bug. As long as `--staged` is broken, the JS script remains as the only enforced check.

Discussed in vault-2025 cleanup session 2026-05-23, see commits `a52bc4c07` (class label recovery) and `a12af5de4` (ABox label recovery) for context — both addressed a related parser bug (label-less class files producing Literal type triples), but `--staged` is a separate orthogonal issue.

## Suggested behavior

When `--staged` is set:
1. Resolve staged file list via `git diff --cached --name-only --diff-filter=ACM`
2. Filter validation to violations whose `focusNode` (or equivalent source IRI) maps to a path in that list
3. Report `violationCount` over filtered set only
4. Exit code reflects filtered conformance

Open to alternative semantics (e.g. validate only graph reachable from staged files), but current "ignore the flag" behavior is unambiguously a bug.

[...]

Issue #2997, opened 2026-05-01
fix(cli): SPARQL regression of #2292 — `Literals cannot appear in subject position` on FILTER(CONTAINS(STR(?var), …))


kitelev:
## Summary

`npx @kitelev/exocortex-cli@15.29.0 sparql query` fails during algebra translation with `❌ Error: Literals cannot appear in subject position` on a query that uses `STR(?var)` to coerce IRI vars to strings inside `FILTER(CONTAINS(...))` / `FILTER(STRENDS(...))`. This is exactly the shape that #2292 was supposed to fix (closed, released via #2307). Issue is reproducing today on production vault.

## Reproduction

CLI: `@kitelev/exocortex-cli@15.29.0`. Vault: `/Users/kitelev/vault-2025` (129 173 triples).

```sparql
PREFIX exo: <https://exocortex.my/ontology/exo#>
PREFIX ems: <https://exocortex.my/ontology/ems#>
SELECT DISTINCT ?task ?label ?status ?updated ?phase WHERE {
  ?task ems:Effort_parent ?phase .
  ?phase ems:Effort_parent ?root .
  ?task exo:Instance_class ?cls .
  ?task exo:Asset_label ?label .
  OPTIONAL { ?task ems:Effort_status ?status }
  OPTIONAL { ?task exo:Asset_updatedAt ?updated }
  FILTER(STRENDS(STR(?cls), '#Task'))
  FILTER(CONTAINS(STR(?root), '5b4030aa-f0c1-43dc-996a-896b0a1a6dfb'))
}
ORDER BY ?phase ?label
```

Output:
```
✅ Loaded 129173 triples in 318135ms
🔍 Parsing SPARQL query...
🔄 Translating to algebra...
❌ Error: Literals cannot appear in subject position
```

## Likely root cause: bad data in graph (cf. #2840)

Loader emits 11 skip warnings during load (these may indicate partial-state inserts that later trip the algebra translator):

```
⚠️ Skipping file with invalid IRI: 03 Knowledge/exo/d4c91968-88a3-46e8-9b6a-9e829248c90e.md — Invalid IRI format
⚠️ Skipping file with invalid IRI: 03 Knowledge/inbox/02c9c7f8-e17a-4fb4-bf0e-8f6338e57c09.md — Literal value cannot be empty
[...]

PR #4009, opened 2026-07-30
feat(cli): opt-in `create --validate` — pre-write SHACL-lite conformance gate (W3)

kitelev:
## What

Adds an **opt-in `--validate` flag to `exocortex-cli create`** that runs the SAME SHACL-lite
conformance check the Write-path PreToolUse hook runs — **before** the write and **scoped to the
new asset**. A non-conformant candidate is refused and **no file is created**.

Project `38800c80` **W3** (RFC `7c7859d1`) — "Опц. inline-SHACL `--validate` в `create` (закрыть
последний hook-gap CLI-пути)".

## Why

A CLI `create` runs through Bash, so the PreToolUse Write/Edit hooks never fire. `create` already
re-implements **three of the four** guarantees those hooks give:

| Guarantee | Write path (hook) | CLI `create` |
|---|---|---|
| Co-location by `exo__Asset_isDefinedBy` | `co-location-enforce.sh` | ✅ #3520 / #3934 |
| Dangling wikilink **VALUE** rejection | `validate-wikilinks.sh` | ✅ `WikilinkValidator` |
| Unknown property **NAME** rejection | — | ✅ RFC 430e84f1 (req `40a9a81b`) |
| **SHACL-lite conformance** | `validate-asset.sh` | ❌ → ✅ **this PR** |

The `ShapeRegistry` `create` already loads was used **only** for cardinality-aware property
*serialization* (#3099/#3179), never for validation. This closes the gap.

## Design

- **Reuse, do not re-implement.** The verdict comes from the very functions
  `validate schema --shapes-mode` uses (`runShapesValidation` = `ShapeLoader.loadFromRDFGraph` +
  `TripleClassHierarchy`, then `applyLegacyExceptionFilter`) → the CLI verdict is *by construction*
  identical to the hook's, and a shapes-semantics change lands in both at once.
- **Pre-write, not write-then-rollback.** The candidate is built in memory (`buildAsset` — the exact
  bytes a write would produce), parsed with the SAME tolerant YAML parser
  `FileSystemVaultAdapter.getFrontmatter` uses, and converted through the production
  `convertNoteFromFrontmatter`. The uid + clock are **pinned first**, so the bytes validated are the
  bytes written (`dry-run-preview-not-real-output`). Strictly better than the hook, which must
  materialise a temp file inside the vault to get its verdict.
- **Candidate-scoped** (load-bearing): without the focus-node filter every create would fail on any
  vault that already has violations elsewhere. Mirrors the hook's temp-file filter (#3245).
- **Severity parity**: only `sh:Violation` gates; `sh:Warning` (open-world unresolvable /
  cross-vault / symbolic refs, #3488) is surfaced on stderr but never blocks.
- **Opt-in, default OFF**: the validator is **lazily imported** inside the flag branch, so the
  default path neither loads the shapes/SPARQL module graph nor changes behaviour. Every validation
  message goes to **stderr** so stdout stays exactly one `{uuid,path,label}` JSON document.
- Composes with `--dry-run` (validates the preview, writes nothing either way).

## Revert-verify

Type-preserving Edit-break of the gate (`throw new ShaclConformanceError(...)` →
`void new ShaclConformanceError(...)`):

```
✕ --validate REFUSES a non-conformant candidate BEFORE writing        ← RED
✓ --validate ACCEPTS a conformant candidate (vault has violations)    ← GREEN (control)
✓ WITHOUT --validate the gate never runs (default OFF)                ← GREEN (control)
✕ --dry-run --validate refuses a non-conformant candidate             ← RED
✓ --dry-run --validate passes a conformant candidate                  ← GREEN (control)
✓ the validated bytes are the bytes written (uid/timestamps pinned)   ← GREEN (control)
Tests: 2 failed, 4 passed
```

Restored → **6/6 GREEN**. Exactly the 2 gate scenarios flip; the 4 negative controls hold in both
states, proving the delta is the candidate-scoped gate and not a blanket refusal.

## Real-vault dogfood (vault-exodev, ~140k triples, `--dry-run` → nothing written)

**Conformant** → `exit 0`, stdout is one clean JSON line, **zero false positives** despite the full
production vault:

```
{"uuid":"743849d9-…","path":"assetspaces/kitelev/exoas-exodev/exodev/743849d9-….md",…}
```

**Non-conformant** (`ems__Effort_status` pointing at an `ems__Project` instead of an
`ems__EffortStatus`) → `exit 2`, **nothing written**:

```
❌ ShaclConformanceError: SHACL-lite validation failed for the new asset — nothing was written.
1 violation(s) in assetspaces/kitelev/exoas-exodev/exodev/b6ac8002-….md:
  • ems__Effort_status (class): sh:class violation: … does not conform to expected class
    https://exocortex.my/ontology/ems#EffortStatus
📋 Context:
  violations: [{"propertyPath":"ems__Effort_status","propertyIri":"…#Effort_status", …}]
```

The error names the **frontmatter KEY** the caller typed (`ems__Effort_status`), not the internal
IRI, plus a machine-readable `{ path, violations }` context an agent can auto-correct from —
exactly the class of typo an LLM agent makes. Verified afterwards: **0 files on disk** for both
candidate uids, vault SHACL still `conforms: true` / 0 violations.

## Test / gate status

- New `@req` integration test: **6/6** (production-shape — real `createCommand()` over a temp vault
  with genuine shapes **and a pre-existing violating asset**).
- Full `packages/cli` suite: **1890 passed**, 1 pre-existing failure
  (`sparql-exo003.integration.test.ts` — fails identically on origin/main with my changes stashed;
  causally isolated).
- `npm run check:types` → **0 errors**. `npm run lint` → clean.
- `lint`-job guard scripts: shard-assignments ✅ · test-antipatterns ratchet ✅ (`55/55`, no
  recurrence) · coverage-monotonic ✅.
- Bumped the brittle `create.test.ts` option count 14 → 15 (additive) + added a `--validate`
  registration assertion.
- `packages/cli/dist/index.js` deliberately **not** committed (vestigial tracked artifact rebuilt
  fresh by CI/`prepublishOnly`; committing a local rebuild is pure minification churn).

Req: dd9ab956-80c7-45c6-b71b-5834f50ce6f8

### DIFF
--- a/packages/cli/src/commands/create.ts
+++ b/packages/cli/src/commands/create.ts
@@ -6,6 +6,8 @@ import {
   ShapeLoader,
   ShapeRegistry,
   GenericAssetCreationService,
+  liveClock,
+  liveUidGenerator,
   type GenericAssetCreationConfig,
 } from "@kitelev/exocortex-core";
 import { NodeFsAdapter } from "../adapters/NodeFsAdapter.js";
@@ -15,7 +17,10 @@ import { WikilinkValidator } from "../services/WikilinkValidator.js";
 import { PropertyNameValidator } from "../services/PropertyNameValidator.js";
 import { EffortStatusResolver } from "../services/EffortStatusResolver.js";
 import { ErrorHandler } from "../utils/ErrorHandler.js";
-import { VaultNotFoundError } from "../utils/errors/index.js";
+import {
+  ShaclConformanceError,
+  VaultNotFoundError,
+} from "../utils/errors/index.js";
 import { registerOrderSpecFromVault } from "../services/registerOrderSpec.js";
 import {
   resolveCoLocationFolder,
@@ -76,6 +81,11 @@ interface CreateCommandOptions {
   yes?: boolean;
   timezone?: string;
   skipWikilinkValidation?: boolean;
+  /**
+   * Opt-in SHACL-lite conformance gate (project 38800c80 W3). Default OFF —
+   * without it `create` is byte-identical (no vault load, no extra output).
+   */
+  validate?: boolean;
 }
 
 /**
@@ -252,6 +262,10 @@ export function createCommand(): Command {
     .option("--yes", "Accepted for symmetry with the apply subcommands (create is non-interactive; no-op)")
     .option("--timezone <tz>", "Timezone for timestamps (defaults to Asia/Almaty)")
     .option("--skip-wikilink-validation", "Skip wikilink existence validation")
+    .option(
+      "--validate",
+      "Run SHACL-lite conformance validation on the new asset BEFORE writing it; a non-conformant asset is refused and no file is created (same shapes as `validate schema --shapes-mode`). Opt-in: omit the flag and create behaves exactly as before.",
+    )
     .action(async (options: CreateCommandOptions) => {
       try {
         const vaultPath = resolve(options.vault);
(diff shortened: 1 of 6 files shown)


kitelev:
## code-reviewer round-1 → applied

**Verdict: APPROVE** — 0 CRITICAL, 0 HIGH, 1 MEDIUM, 3 LOW. The reviewer independently re-ran the
revert-verify (2 RED / 4 GREEN, restored byte-identically) and independently verified the
`sparql-exo003` failure is pre-existing (pristine origin/main worktree, fails identically). It also
**refuted its own suspicion** on the riskiest axis: the focus-node round-trip (`encodeURI` on emit
vs `decodeURIComponent` on decode) was probed with adversarial paths — spaces, Cyrillic, `#`, `%`,
`[]` — all MATCH, so the feared silent fail-open does **not** occur.

### MEDIUM — fixed in c63330f7

> `--validate` crashes with an opaque error on a present-but-empty property value.

Real defect, and it defeated the requirement's machine-readable-error clause for a reachable input.
Root cause: the gate called `convertNoteFromFrontmatter` **directly**, bypassing the two-phase
commit `convertVault` applies (#2997 Phase 2) — it validates file-level invariants *before*
converting and SKIPS any file that fails one. The candidate therefore took a code path a real
indexed file never reaches.

Fixed by restoring parity in **both** halves: a `validateExocortexAsset` pre-check (which also fails
**fast**, before the vault-wide load) plus a try/catch backstop around the conversion. An
unindexable candidate is reported as a **gate violation**, not silently passed — it emits ZERO
triples, so the asset would be invisible to every query, layout and graph relation: the exact
"silently dead asset" class this workstream exists to prevent, and strictly worse than a shape
violation.

Real-vault verification (vault-exodev, `--dry-run`):

```
EXIT=2   stdout: []
❌ ShaclConformanceError: SHACL-lite validation failed for the new asset — nothing was written.
  • ems__Effort_status (EMPTY_OPTIONAL_PROPERTY): Optional property "ems__Effort_status" is
    present but empty — the asset would be SKIPPED by the indexer (no triples emitted),
    making it invisible to every query, layout and graph relation.
```

(previously: a raw `Literal value cannot be empty`).

### LOW — dispositions

- **LOW 1 (parity docstring stronger than reality)** — **fixed** in the same commit: the docstring
  now names the deliberate SKIP-case divergence.
- **LOW 2 (no `--use-cache` parity)** — **kept as-is, deliberate.** Cache staleness in a *gate* is
  worse than a cold load; the flag is opt-in. Documented trade-off.
- **LOW 3 (brittle option-count assertion)** — pre-existing; this PR only bumps 14 → 15 and **adds**
  a behavioural `--validate` registration assertion beside it. Ratchet clean (`method-exists=55/55`,
  no recurrence).

### Gate status after the fix

- `@req` integration test **7/7** (a 7th case now pins the unindexable path).
- Revert-verify re-run: neutralising the gate reds exactly the **3** gate scenarios (2 shape +
  1 indexability); the **4** negative controls stay GREEN → still non-vacuous.
- Full `packages/cli`: **1891 passed**, 1 pre-existing failure (`sparql-exo003`).
- `check:types` 0 errors · anti-pattern ratchet ✅ · shard-assignments ✅ · coverage-monotonic ✅.

PR #4010, opened 2026-07-30
feat(cli): MCP-MVP — `exocortex mcp` stdio server (create_asset / create_effort + inline-SHACL)

kitelev:
## What

Adds `exocortex mcp` — an **MCP (Model Context Protocol) stdio server** exposing two asset-creation tools, as a CLI **transport** entry-point.

- `create_asset` → the generic `create` verb (inline-SHACL `--validate` **ON by default**)
- `create_effort` → `apply <cmd> <target> --json --yes` (the *homoiconic* effort path)

Project `38800c80` **W6**, RFC `7c7859d1`.
Req: `264ccef6-4737-4076-95d8-053326c9ee1d` (approved).

## Why

The dogfood-canonical creation path is invoked through a **shell**, and shell quoting is the most-recurring mechanical failure class in this environment: word-splitting of multi-word variables, `$`/backtick command substitution inside `--label`, `()` parse errors, glob expansion of unquoted `?`/`*`. Those are **transport** defects, not pipeline defects.

An MCP client hands over a JSON object; the server builds an **argument vector** and runs this same CLI with `execFile` (**no shell**), so no character in any label, body or property value can be reinterpreted.

## Design

- **Purely additive.** New `packages/cli/src/mcp/**` + `commands/mcp.ts` + 2 lines in `program.ts`. **Zero changes to `create.ts` / `apply.ts`**, so regression risk on the creation path is zero.
- **Zero new runtime dependencies.** The required MCP subset (`initialize`, `notifications/initialized`, `ping`, `tools/list`, `tools/call`) is a small JSON-RPC 2.0 dialogue over newline-delimited JSON — hand-rolled, leaving a *published* package's supply chain unchanged. Adopting the official SDK is a documented follow-up once value is proven.
- **Inherits every existing guarantee** by shelling out to the real CLI: co-location by `exo__Asset_isDefinedBy`, dangling-wikilink VALUE rejection, unknown-property-NAME rejection (req `40a9a81b`), SHACL-lite conformance (req `dd9ab956`), canonical property ordering.
- **inline-SHACL ON by default** on the MCP surface (the CLI keeps `--validate` opt-in for interactive use); overridable per call.
- **A failing CLI run is an MCP tool error, never a crash** — non-zero exit → `isError: true` carrying the CLI's own stderr diagnostic + exit code.
- `mcp` is a **transport** entry-point, not a domain verb — it does not count against the RFC's 22→18→17 domain-command surface metric.

## Verification

**Revert-verified** (`@req:264ccef6-4737-4076-95d8-053326c9ee1d`), each break reddening **exactly** its own test with the other 11 green:

| break | RED test |
|---|---|
| route `execFile` through a shell (`shell: true`) | byte-identical-argv |
| `validate !== false` → `validate === true` | inline-SHACL default-ON |
| neutralise the non-zero-exit branch | CLI-failure → tool error |

Restored → **12/12 GREEN**.

**End-to-end against the built CLI** (real MCP stdio session, real vault):
- handshake / `tools/list` / `tools/call` all correct; notification correctly **unanswered**
- a real asset was created whose hostile label `$(whoami) \`id\` $HOME $& (paren) "q" 100%` and body (`; echo pwned`) are **byte-identical on disk**, with **no shell side-effects**
- unknown tool → `isError`; unsupported method → `-32601`
- `create_effort` against a bogus command → `isError` with the CLI diagnostic, server survived (exit 0)

**Gates:** `check:types` 0 errors · CLI suite 153/154 suites, 1903 passed · archgate 24/24 pass, 0 errors · all 3 `lint`-job guard scripts exit 0 (ratchet unchanged: method-exists 55/55, vacuous-length 21/21).

> The single failing suite (`sparql-exo003.integration.test.ts`) is **pre-existing on `origin/main`** — independently reproduced in a clean worktree containing zero MCP code.

## Out of scope (documented, not dropped)

`@modelcontextprotocol/sdk` adoption · MCP `resources`/`prompts` capabilities · read/query tools (RFC read-axis = W5, deferred) · making MCP mandatory (W8 hard-hook, gated on the W7 metric).

### DIFF
--- a/packages/cli/src/commands/mcp.ts
+++ b/packages/cli/src/commands/mcp.ts
@@ -1,0 +1,85 @@
+import { Command } from "commander";
+import { createInterface } from "readline";
+import {
+  McpServer,
+  JSON_RPC_PARSE_ERROR,
+  type JsonRpcMessage,
+  type JsonRpcResponse,
+} from "../mcp/McpServer.js";
+import { createSelfCliRunner } from "../mcp/cliRunner.js";
+
+interface McpCommandOptions {
+  entry?: string;
+}
+
+/**
+ * Drive an {@link McpServer} over newline-delimited JSON on stdio.
+ *
+ * Exported for testing: the caller supplies the streams, so the loop can be
+ * exercised without spawning a process.
+ */
+export async function runStdioServer(
+  server: McpServer,
+  input: NodeJS.ReadableStream,
+  write: (line: string) => void,
+): Promise<void> {
+  const rl = createInterface({ input, crlfDelay: Infinity });
+
+  for await (const rawLine of rl) {
+    const line = rawLine.trim();
+    if (line.length === 0) {
+      continue;
+    }
+
+    let message: JsonRpcMessage;
+    try {
+      message = JSON.parse(line) as JsonRpcMessage;
+    } catch {
+      const parseError: JsonRpcResponse = {
+        jsonrpc: "2.0",
+        id: null,
+        error: {
+          code: JSON_RPC_PARSE_ERROR,
+          message: "Parse error: invalid JSON",
+        },
+      };
+      write(JSON.stringify(parseError) + "\n");
+      continue;
+    }
+
+    const response = await server.handle(message);
+    if (response !== null) {
+      write(JSON.stringify(response) + "\n");
+    }
+  }
+}
+
+/**
+ * `exocortex mcp` — MCP (Model Context Protocol) stdio server exposing the
+ * Exocortex asset-creation tools (project 38800c80 W6, req 264ccef6).
+ *
+ * This is a TRANSPORT entry-point, not a domain verb: it adds no new creation
+ * semantics. Each tool re-invokes this same CLI with an explicit argument
+ * vector (no shell), so every guarantee of the creation path is inherited and
+ * no caller-supplied value can be reinterpreted by a shell.
+ */
+export function mcpCommand(version?: string): Command {
+  return new Command("mcp")
+    .description(
+      "Run an MCP (Model Context Protocol) stdio server exposing create_asset / create_effort tools backed by this CLI",
+    )
+    .option(
+      "--entry <path>",
+      "Path to the CLI entry script the tools invoke (defaults to this process's script)",
+    )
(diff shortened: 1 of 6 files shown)


kitelev:
### Review response — both HIGH findings fixed in `cfae88bc`

Thanks — both were real, and the second one (`--aliases`) is the interesting one: `execFile` closes the **shell** layer, but the CLI's **own argument parser** is a second interpretation layer the original design didn't account for. Good catch.

| # | Finding | Fix |
|---|---|---|
| **HIGH-1** | variadic `--aliases` lets an alias inject real CLI flags (`--vault`, `--skip-wikilink-validation`, `--body-file`, `--created-by`) | one `--aliases` per value, mirroring what the `--property` loop already did |
| **HIGH-2** | a bare `null` line throws past the parse guard and kills the long-lived loop | non-object guard in `handle()` **+** try/catch in the stdio loop → `-32603` |
| **MEDIUM-1** | a throwing `runCli` escapes `handle()` | covered by the same try/catch (the load-bearing half — makes design point 4 true *by construction*, not per-input) |
| **MEDIUM-2** | non-string property values → `[object Object]`; `aliases: "str"` spreads per character | `invalidCreateAssetShape()` fails loud naming the offending field |

**HIGH-1 verified empirically against the real Commander parser**, to the same standard you used to prove the vulnerability — the fixed argv resolves the caller's vault:

```
vault the CLI actually resolved -> /tmp/LEGIT
redirected to ATTACKER?          NO
```

**Revert-verified** (two new axes, same style as the original three):

| break | RED test |
|---|---|
| restore the spread `--aliases` emission | alias-injection |
| neutralise the loop try/catch | loop-survivability |

Each reddens exactly its own test; restored → **15/15 green**. Full CLI suite 1906 passed, `check:types` 0, archgate 24/24 / 0 errors, all 3 `lint`-job guards exit 0.

**LOW items** deliberately not taken: `parseLastJsonLine` whole-stdout fallback (both current emitters are single-line; noted for if `create` ever pretty-prints), the `"Unknown tool: undefined"` wording, and stdout backpressure (irrelevant at MCP payload sizes).

PR #3938, opened 2026-07-25
chore(deps): npm audit fix — resolve high advisories (brace-expansion/qs/uuid) blocking Auto Release

kitelev:
## What

Clears all `npm audit --audit-level=high` findings (**39 → 0**), which gate Auto Release (the `npm-audit-exocortex` / `NPM Security Audit` jobs → the full CI `workflow_run.conclusion`).

## Root advisories (freshly-published GHSA — advisory-drift recurrence)

`main`'s own CI run predates these advisories (so it's green); the **next** release-gate run would fail (`npm audit` runs against the live advisory DB).

| advisory | severity | note |
|---|---|---|
| `brace-expansion` ≤5.0.7 (GHSA-mh99-v99m-4gvg, DoS) | 36 high | cascades through the dev toolchain via `minimatch` (eslint / jest / typescript-eslint) |
| `qs` 6.11.1–6.15.1 (GHSA-q8mj-m7cp-5q26, DoS) | high | |
| `uuid` 13.0.0 (GHSA-w5hq-g745-h8pq, buffer bounds) | moderate | the only **runtime** dep |

## Why this shape (not `audit fix --force`)

- `npm audit fix --force` **could not resolve it** (`fixAvailable: false` — the dev-tools' pinned old `minimatch` transitively pins vulnerable `brace-expansion`; no auto path).
- Root `overrides` alone left residuals — **npm 11 does not force a workspace-nested `minimatch`**. Traced the last holdout: `packages/cli → glob@10 → minimatch@9 → brace-expansion@2.1.2`. `glob@11` uses `minimatch@10` (→ patched `brace-expansion@5.x`).

## Changes

- root `overrides`: `minimatch ^10.0.3`, `brace-expansion 5.0.8`, `qs ^6.15.3`
- `uuid` dependency `^13.0.0` → `^13.0.1`
- `packages/cli` `glob` `^10.3.0` → `^11.0.0` (clears the last `minimatch@9` holdout)

`npm audit --audit-level=high` → **0 vulnerabilities**.

## Toolchain validated (the minimatch 3→10 / glob 10→11 glob-behavior risk did NOT materialize)

`check:types` ✓ · `lint` ✓ (eslint minimatch file-matching) · `build` ✓ (core+services+plugin+cli) · jest core smoke ✓ · **cli integration 5 suites / 74 tests** ✓ (glob 11 direct).

https://claude.ai/code/session_01Lzk1cejAQeEPsj6GtFQrRo

PR #4034, opened 2026-08-07
chore(deps): unblock release gate — GHSA-5p4m-2wfm-xmqj

kitelev:
## Why (release gate is red repo-wide)

`main` fails the **non-required** job `npm-audit-exocortex`. That reddens `CI.workflow_run.conclusion`, and Auto Release gates on `workflow_run.conclusion == 'success'` — so **the release has been silently SKIPPED for every merge**, repo-wide. `d75eaa07` (merged 2026-08-07) never shipped to the user.

This is the failure mode documented in `rules/auto-release-gated-on-ci-workflow-conclusion.md`: required-14 checks let PRs merge, but a non-required job still blocks the release.

## Root cause (verified, not inferred)

| | |
|---|---|
| Gate command | `npm audit --audit-level=high --omit=dev` |
| Gate cwd | `packages/core` |
| Advisory | **GHSA-5p4m-2wfm-xmqj** (high, prototype pollution), `js-yaml >=4.0.0 <4.3.1`, fixed in **4.3.1** |

⛤ **The audit input is the monorepo ROOT `package-lock.json`, not `packages/core/package-lock.json`.** `packages/core` is an npm **workspace member**, so npm walks up to the root lockfile. The `cache-dependency-path: packages/core/package-lock.json` line in `ci.yml` is only a **setup-node cache key** — conflating it with the audit input sends you to the wrong lockfile (I went there first).

In the root lockfile `node_modules/js-yaml` was pinned at **4.3.0** with `dev` unset ⇒ **production path** ⇒ `--omit=dev` correctly flagged it.

The gate already scopes to `--omit=dev` (shipped 2026-07-27), so no gate change was needed — the exposure is genuinely production.

## Fix (split-exposure — targeted consumer bump, no global override)

`packages/cli` is the **only PUBLISHED prod consumer** of js-yaml. Verified against `origin/main` manifests:

| package | js-yaml declared as | in `--omit=dev` prod graph? |
|---|---|---|
| `packages/cli` | `dependencies` | ✅ yes — and it is the **published** one |
| `packages/req-audit` | `dependencies` | ✅ yes (see Observation 1 — resolves to the same hoisted node) |
| `packages/core` | devDependencies | no |
| `packages/obsidian-plugin` | devDependencies | no |

⚠ Both `cli` and `req-audit` sit in the prod graph; `private: true` does **not** remove a workspace member's `dependencies` from it. Pinning `cli` suffices only because both resolve to the single hoisted `node_modules/js-yaml` — see Observation 1.

Its range `^4.2.0` **already permitted 4.3.1** — only the lockfile held the vulnerable resolution.

- `packages/cli/package.json`: `js-yaml ^4.2.0 → ^4.3.1` (same major, patch)
- `package-lock.json`: `npm audit fix --package-lock-only` (**non-force**)

⛔ Deliberately **no global `overrides`** — that path was tried in closed PR #3938 and broke the jest coverage pipeline (`minimatch@10` removed the callable default export that babel-plugin-istanbul still calls). Bumping the shipped consumer's own dep sidesteps it entirely.

### Lockfile delta

Net: `js-yaml 4.3.0 → 4.3.1` on the prod path. Incidental **patch-level** bumps of dev-only transitives: `brace-expansion` (1.1.16→1.1.18 ×6, 2.1.2→2.1.4), `js-yaml@3` under `@istanbuljs/load-nyc-config` (3.15.0→3.15.1), `fast-uri` 3.1.4→3.1.5, `nanoid` 3.3.16→3.3.17, `postcss` 8.5.21→8.5.26. **No major bumps.**

## Verification

Per `rules/auto-release-gated-on-ci-workflow-conclusion.md` §2026-07-25 — validate the packages that **USE** the bumped dep, not a core-only smoke (a core-only smoke is exactly what let PR #3938 look safe while breaking every coverage shard).

| check | result |
|---|---|
| `npm audit --audit-level=high --omit=dev` (cwd `packages/core`) | **exit 0** — `found 0 vulnerabilities` |
| `npm ci` | exit 0, 951 packages (lockfile ↔ manifest in sync) |
| `npm run check:types` | exit 0 |
| core js-yaml-heavy suites `--coverage` | **38 suites / 617 tests** pass |
| obsidian-plugin suites `--coverage` | **11 suites / 119 tests** pass |
| cli full suite | 155/156 suites, 1862 tests pass |

The single cli failure (`sparql-exo003.integration.test.ts`) is **pre-existing and local-environment-only** — proven by a flip: `npm install js-yaml@4.3.0 --no-save` reproduces the *identical* failure, and `test-coverage-cli` is green on `main`. Not caused by this change; not investigated further (out of scope).

⚠ `packages/exoas-exo` had to be `git submodule update --init` — 3 plugin parity tests fail on an uninitialised submodule. Environmental, unrelated to this diff.

## Observations (NOT actioned here — deliberately out of scope)

> **Corrected after orchestrator review** — my first draft of both items below stated a *false mechanism*. Both are now re-derived from the lockfile diff in this PR and from `gh api advisories`, not from plausibility. Recording the correction rather than silently overwriting, because a wrong-but-true-sounding mechanism in a durable artifact is worse than no note (the next reader treats it as verified).

### 1. `packages/req-audit` also declares js-yaml in `dependencies` — and that is fine, but NOT for the reason I first wrote

⛔ **Wrong reason (my first draft):** "it is `private: true`, so it is not shipped and the gate does not require it."

`private: true` governs **publication to npm**, not **classification in the audit graph**. A workspace member's `dependencies` land in npm's **production** graph regardless of `private` — so `--omit=dev` *does* see them. Left as written, that note would teach the next reader that private packages are exempt from the audit gate. They are not.

✅ **Actual mechanism:** the root lock contains exactly **two** js-yaml nodes —

```
node_modules/js-yaml                                        4.3.1  dev=None   <- hoisted, prod
node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml 3.15.1 dev=True
```

`packages/req-audit` has **no nested `node_modules` at all**, so it resolves to that same hoisted node — which this PR already moved to 4.3.1. Nothing to do, because it is *already covered*, not because it is exempt.

### 2. `brace-expansion` — the July override does not merely lag, it BLOCKS the patch

**GHSA-rgw5-rvv9-x895** (published 2026-08-03, high) patches four lines:

| vulnerable range | patched |
|---|---|
| `< 1.1.18` | 1.1.18 |
| `>= 2.0.0, < 2.1.4` | 2.1.4 |
| `>= 3.0.0, < 3.0.6` | 3.0.6 |
| `>= 4.0.0, < 5.0.9` | **5.0.9** |

(My first draft said the range topped out at 5.0.8. It does not — the patch is **5.0.9**, and 5.0.9 is published on npm.)

⛤ **The lockfile diff in this PR is itself the evidence.** `npm audit fix` moved every brace-expansion node it was *permitted* to move, straight to this advisory's patched versions:

```
1.1.16 -> 1.1.18   x6 nodes   dev=True    (no override on 1.x -> floated freely)
2.1.2  -> 2.1.4    x1 node    dev=True    (no override on 2.x -> floated freely)
5.0.8  -> 5.0.8    x1 node    dev=None    <- PINNED by overrides["brace-expansion@5.x"]="5.0.8"
```

The one node left vulnerable is the **production** one (`cli -> glob@11 -> minimatch@10.2.5 -> brace-expansion`, `dev` unset), and it is stuck precisely because the July override pins it *below* the 5.0.9 patch. So the override is now actively preventing the fix npm would otherwise have applied by itself.

`npm audit --omit=dev` does **not** flag it today, and did not flag it before this change either (the failing CI log showed exactly one high = js-yaml) — so this PR is not hiding a regression. But it is a live high-severity node on the prod path and a strong candidate for the next gate-reddening event.

**Not touched here on purpose** — bumping the override collides with this PR on the root lockfile. The orchestrator is taking it as a separate work-item after this merge.

## Definition of Done

- [x] `npm audit --audit-level=high --omit=dev` = 0
- [x] `git diff --stat` = only the 2 expected files
- [x] `check:types` + coverage smoke green on **consumer** packages
- [ ] push-run CI on `main` = `conclusion: success`
- [ ] Auto Release for that run = `completed/success`, new tag containing `d75eaa07`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2559,9 +2559,9 @@
       }
     },
     "node_modules/@eslint/config-array/node_modules/brace-expansion": {
-      "version": "1.1.16",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.16.tgz",
-      "integrity": "sha512-IDw48K2/2kRkg9LdJxurvq3lV3aBgq0REY89duEqFRthjlPdXHKMj7EnQOXVckxzgisinf3nHfrcE2FufFLXMw==",
+      "version": "1.1.18",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.18.tgz",
+      "integrity": "sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -2650,9 +2650,9 @@
       }
     },
     "node_modules/@eslint/eslintrc/node_modules/brace-expansion": {
-      "version": "1.1.16",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.16.tgz",
-      "integrity": "sha512-IDw48K2/2kRkg9LdJxurvq3lV3aBgq0REY89duEqFRthjlPdXHKMj7EnQOXVckxzgisinf3nHfrcE2FufFLXMw==",
+      "version": "1.1.18",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.18.tgz",
+      "integrity": "sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -2909,9 +2909,9 @@
       }
     },
     "node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml": {
-      "version": "3.15.0",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.15.0.tgz",
-      "integrity": "sha512-ttBQIIQPDeLjpPOohtUdXuXUVoA2uIB6fEH9HyJ7234s5mBJ5wTx20njxplLZQgLaOfpmPQA7X2t5AX6tIPbog==",
+      "version": "3.15.1",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.15.1.tgz",
+      "integrity": "sha512-S99WuO3HlhO3XN41EtYUNl9zzXjoJx7QvmipxsJVxtCBT0YHEFy+iOJhjSvrmV12nYhWpZaM8lPHkJm0yUMbag==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -5575,9 +5575,9 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.2.tgz",
-      "integrity": "sha512-w5JZcKgdhDOgOwm8H+KgbosopHMuGcl6qbulwjtz3SM7I7P3yW1eAjzMPLrIE+NQ9vjgANKHWeMHnrT0OXW1oA==",
+      "version": "2.1.4",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.4.tgz",
+      "integrity": "sha512-hGfVzPxthbf3+2yjg/RBs60cB0FhqBS/zvdV/4wn4/BmN0bNMMHPc4V/BbFieqf1TKAGGAHnY4eSjajCl0f2Xg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


kitelev:
### Why the existing Dependabot PR did not (and could not) fix this

**#4028** ("Bump the npm_and_yarn group across 1 directory with 2 updates", open since **2026-08-04**) touches:

```
packages/core/package-lock.json
packages/core/package.json
```

That is the **wrong lockfile for this gate**. `packages/core` is an npm **workspace member**, so `npm audit` invoked from `packages/core` resolves against the **root** `package-lock.json` — `packages/core/package-lock.json` is not the audit input. (The `cache-dependency-path: packages/core/package-lock.json` line in `ci.yml` is only a `setup-node` **cache key**, which is what makes this so easy to conflate — I went down that path first and regenerated the core lock, which produced *literally zero* changes. That zero-delta is what exposed the mistake.)

So #4028 would have merged, looked plausible, and left the release gate **still red** — which is consistent with what actually happened: the gate has been red and Auto Release skipped for every merge since, with #4028 sitting open the whole time.

**Follow-up candidate (not actioned here):** Dependabot appears to be configured against a workspace-member lockfile rather than the root one for this ecosystem, so its npm PRs for this monorepo may be systematically ineffective at moving the audit gate. Worth checking `.github/dependabot.yml` `directory:` config separately.

### Baseline this PR is measured against

| | |
|---|---|
| CI push-run for `d75eaa07` | [`31140597547`](https://github.com/kitelev/exocortex/actions/runs/31140597547) — `completed/failure` |
| Sole failing job | `npm-audit-exocortex` (non-required) |
| Auto Release runs for `d75eaa07` | `31141085874`, `31140808288` — **both `completed/skipped`** |
| Latest release | `v16.213.1` (2026-08-06T13:36) — nothing shipped since |

### Predictive check on this PR head

`npm-audit-exocortex` on `5fc1dd5f`: **`completed/success`**.

⛤ Worth stating explicitly: the required-14 checks would **never** have shown this — `npm-audit-exocortex` is *non-required*, which is the whole reason a red gate still allowed merges while silently killing the release. The required-set going green is necessary but not sufficient evidence for this PR; the non-required job passing is the decisive one.

PR #4035, opened 2026-08-07
test(sync): lock the FINDING-3 mount-walk guards individually and as a pair

kitelev:
## What

Locks the two guards inside `enumerateMountFolders` (`SyncDepsFactory.ts`) that jointly hold one contract stated in its own docstring — «a detection helper must never throw into the sync path»:

- **B2** — `if (!(await adapter.exists(root))) return out;` (absent-root fast path)
- **C** — the outer `catch` that swallows any listing failure

**Test-only.** `SyncDepsFactory.ts` is byte-identical to `origin/main` (`git diff origin/main...HEAD -- <prod file>` = 0 lines).

## Measured (mutation on a hardlink copy of `origin/main@0cc49ef2`)

One mutation per guard, exact-literal anchor, **replacement count asserted** so a mutation that silently fails to apply cannot pass for green. The working tree is never touched (the mutated file's hardlink is broken first; the driver asserts `st_ino` differs).

| mutation | red before | red after |
|---|---|---|
| B2 alone | 0 | 0 |
| C alone | 0 | **1** |
| B2 + C together | 3 | 5 |

Two checks added, doing **different** jobs — only the first adds coverage:

- **«isolates the outer swallow»** — a mount root that EXISTS but cannot be listed (EACCES / iOS sandbox) reaches C and nothing else. Turns a previously non-isolable guard into an individually observable one; it is the only red under `C alone`.
- **«pair axis for the absent-root fast path»** — B2 has no isolable behaviour *by construction* (C covers the very same input), so it is locked as a pair. Adds **no** coverage (the pair reds 4x without it). The pair *was* already locked — but only **incidentally**, by three tests whose subject is unrelated («skips non-GitHub sources», «warns on a FileSpace without a source», «excludes AssetSpaces whose mount folder is absent»), each green merely because its fixture happens to omit `assetspaces/`. Adding that folder to any of them would unlock the pair silently. This check states the contract by **name** on a fixture whose root-absence is deliberate — intent, not coverage.

⛔ **Zero-red-alone means the contract is held TWICE, not that it is untested** — neither guard may be deleted as «uncovered» (`integration-test-revert-verify` §ДУБЛИРУЮЩИЕ гварды).

## The task's premise was false — recorded here so the numbers don't outlive it

The task claimed `adapter.exists` and the walk-root scoping were **individually uncovered** (0 red alone). Reproducing the measurement first (rather than trusting it) showed both are **already individually observable — 3 red each**. Cause: the file has **nine** `adapter.exists` call sites, and line 396 carries the same `exists(spec.localPath)` substring as line 132; the earlier measurement anchored on the wrong one. Its «both guards -> 2 red» reproduces exactly as my **A + B2**, which is the fingerprint of that mis-anchor.

So the genuine residual was not the pair the task named, but the **B2 + C** pair — which this PR locks.

| guard | gates | individually observable on `main`? |
|---|---|---|
| `exists(spec.localPath)` L132 | `specs` | yes — 3 red |
| walk-root scoping `const root = "assetspaces"` | `mountedNotDeclared` | yes — 3 red |
| B2 — `exists(root)` early return | fast path | no — 0 (non-isolable by construction) |
| C — outer `catch` | swallow | 0 -> **1 after this PR** |

## No `@req:` tag / no new requirement

This hardens an **existing internal resilience contract** (the helper's own docstring), not a new product invariant — `4eca4900` covers parking, a different invariant. `/feature-sdd` Step 0 exempts test-only hardening of an already-shipped contract.

## Gates

- target suite 26/26 · full plugin suite **338 suites / 6049 passed**
- `check:types` exit 0
- all three CI `lint` guard scripts exit 0 (`check-test-antipatterns.sh` ratchet unchanged: method-exists 55/55, vacuous-length 21/21)
- `code-reviewer`: **APPROVE** — all four claims verified empirically, matrix reproduced cell-for-cell. Its MEDIUM + two LOWs are applied in the second commit (the first commit's docstring overstated «the pair was locked only INCIDENTALLY *until these checks*» — measured, only one check moved coverage; corrected in a follow-up commit rather than silently rewritten).

Vault task: `1d002b6b-fd50-4eeb-abd2-11800fd0e62e`

### DIFF
--- a/packages/obsidian-plugin/tests/unit/sync/SyncDepsFactory.test.ts
+++ b/packages/obsidian-plugin/tests/unit/sync/SyncDepsFactory.test.ts
@@ -177,13 +177,16 @@ describe("collectSyncRepoSpecs", () => {
    *   B2 — `if (!(await adapter.exists(root))) return out;`   (absent-root fast path)
    *   C  — the outer `catch` that swallows any listing failure
    *
-   * Measured on a hardlink copy of origin/main@0cc49ef2, this suite:
+   * Measured by mutating a hardlink copy of origin/main@0cc49ef2 and running
+   * this suite against it — one mutation per guard, exact-literal anchor, with
+   * the replacement count asserted so a mutation that fails to apply cannot
+   * pass for green:
    *
-   *   | mutation        | red |
-   *   |-----------------|-----|
-   *   | B2 alone        |  0  |
-   *   | C  alone        |  0  |
-   *   | B2 + C together |  3  |
+   *   | mutation        | red before | red after |
+   *   |-----------------|------------|-----------|
+   *   | B2 alone        |     0      |     0     |
+   *   | C  alone        |     0      |     1     |
+   *   | B2 + C together |     3      |     5     |
    *
    * ⛔ Zero-red-alone means the contract is held TWICE, NOT that it is
    * untested — neither guard may be deleted as «uncovered»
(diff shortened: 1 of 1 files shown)


kitelev:
## ⛔ Correction — the «No `@req:` tag / no new requirement» section of this PR body is WRONG

Found during this session's retrospective (Step 3a cross-reference), after merge. Recording it here rather than silently, so the false justification does not outlive the PR.

**What the body claimed:**

> This hardens an existing internal resilience contract (the helper's own docstring), not a new product invariant — `4eca4900` covers parking, a different invariant. `/feature-sdd` Step 0 exempts test-only hardening of an already-shipped contract.

**Sentence 1 is correct** — verified by reading `4eca4900` on the reqs remote: its scenarios and `covers` bullets are about *parking* (a declared AssetSpace whose derived mount path is absent must not be enumerated). Guard C's contract — an unlistable root degrades to `[]` instead of throwing into sync (EACCES / iOS sandbox) — is nowhere in it.

**Sentence 2 is false.** `/feature-sdd` Step 0 carries an addendum dated **2026-08-06** whose empirical is **PR #4032 / req `4eca4900`** — the immediately preceding task in this same work stream:

> ⛤ **TEST-ONLY PR, чей продукт — САМА СПЕЦИФИКАЦИЯ ранее-неявного свойства = ПОЛНЫЙ ЦИКЛ, а не «refactor».** … Отнеся это к refactor, ты отгружаешь тест **без владельца**: `requirements-trace` о нём не узнает, и следующий рефактор снимет гвард при зелёном CI.

Its discriminator is mechanical, and **its input is the matrix already published in this PR**:

| kill the guarded mechanism on a copy of `origin/main`, before the new tests | routing |
|---|---|
| something reds | property already specified → extend `@req` of an existing req |
| **zero red** | **property true but unlocked → full loop + NEW req** |

Measured here: **`C alone` → 0 red**. So the correct route was a **new requirement**, not exempt.

**Consequence (real, not cosmetic):** the two tests added by this PR carry no `@req:` tag, so `requirements-trace` does not know they exist. A future refactor can delete them on a green CI — which is precisely the failure mode this PR was written to prevent. Nothing is broken today; the guard-locking test is itself unguarded.

**Cause:** the spawn prompt named the skill imperatively (`Дальше — Skill: feature-sdd как рамка`), pre-empted the objection («дом-процесс для ЛЮБОГО изменения кода, **включая test-only**»), cited the sibling precedent (#4032 → `4eca4900`), and made the decision an explicit step («Реши на STEP B, покрывает ли существующий req новую ось, или нужен свой»). The skill was never opened; the routing was decided from memory at the end of the work, while the discriminator's input had been sitting in the matrix since STEP A.

**Remaining work (not done here — a new req needs Andrey's approval by design):**

1. `Skill: feature-sdd` from Step 1 — author the req for the resilience contract («a detection helper must never throw into the sync path»), Gherkin from the two shipped scenarios (absent root; existing-but-unlistable root), `bindingClass: unit`.
2. Approve gate.
3. Follow-up PR tagging both tests `@req:<new-uid>`, merged into main.
4. **Only then** flip the req to Active — per `req-first-sdd-ci-binding`, flipping before the tags are in main reds the active-gate repo-wide for everyone.

Mitigations applied to the corpus in the same retrospective: `feature-sdd` Step 0 gained the tell «you are justifying exempt with prose while unable to name the red count», and `skill-discipline-under-pressure` gained the recurrence «a *complete and correct* paraphrase suppresses opening the source harder than an incomplete one».

PR #4044, opened 2026-08-13
test(e2e): the PARKED-link placeholder renders in live Obsidian (req c171e24d)

kitelev:
Closes **AC7** of vault task `24993104-c74d-429b-997c-d23e770816b9` — the visual gate for the parked-link placeholder shipped in #4043.

## Why this exists on top of the jsdom tests

The unit tests in `tests/unit/presentation/parked/ParkedLinkPlaceholder.test.ts` drive the real entry points (`BodyLinkPatch.enable()` / `PropertiesLinkPatch.enable()`) and already prove the DOM output of both channels. They cannot be the **visual** guarantee, because they replace with a fake exactly the three things that break in production:

- the `metadataCache` is **cold** on a fresh launch — the defect class that painted the profile indicator `Unknown` forever and that a manual smoke cannot reproduce *in principle* (a live vault's cache is always warm);
- the reading-mode markdown render + MutationObserver are Obsidian's, not jsdom's — whether a `[[uid]]` even *becomes* an `a.internal-link` carrying `data-href` is Obsidian's decision;
- the parked lookup reads a **real dot-folder** through the **real** `vault.adapter`. The unit test's fake reader models the `getMarkdownFiles()`-vs-adapter asymmetry that *is* the whole mechanism; here the filesystem provides it.

## Scenarios (both tagged `@req:c171e24d-15d3-4073-a34b-f6e78d3bc15f`)

1. **body channel** — the parked link renders `«<label> ⏸»`, carries the AssetSpace derived from the real parked path, and keeps that name **out** of the tooltip.
2. **frontmatter channel** — the same placeholder through `PropertiesLinkPatch`, pointing at a **different** parked asset so a cross-channel bleed cannot pass.

## Non-vacuity

A placeholder that swallowed every unresolved link would satisfy a positive-only scenario. The body note therefore carries links whose verdicts must **differ**, and each negative control is asserted to have actually been **looked up** (`data-parked-checked`) rather than merely left alone:

| link | expected |
|---|---|
| parked asset | placeholder |
| UID present **nowhere** | checked, still broken |
| non-asset file under the parked root | checked, still broken (`exo__Asset_uid` gate) |
| asset inside the mount's own `.git` | checked, still broken (nested dot-folder skip) |
| **resolved** link | never checked at all |

## Scope

Test-only; zero production change. Step 0 was routed by a **mutant on `origin/main`**: killing the body-channel `decorate()` call reds 1 of the 12 existing tests, so the property is already specified by req `c171e24d` — this **extends that req's `@req` coverage** rather than minting a new requirement.

## Verification

- `playwright --list` discovers both scenarios with full-UUID `@req` tags.
- The three `lint`-job guard scripts pass; the suite sits outside `tests/e2e/specs`, so `validate-shard-assignments` still sees 21 balanced files and **this PR cannot be blocked by an eka-gui flake** (non-required check).
- Revert-verify runs in CI (local Docker/QEMU is forbidden on this machine): a follow-up push removes the `decorate()` call from `BodyLinkPatch` and the run must red **only** the body scenario, leaving the frontmatter one green — the discriminator the requirement's own Gherkin names.

Req: c171e24d-15d3-4073-a34b-f6e78d3bc15f

https://claude.ai/code/session_014wmg6gVN8uTZEdVN2BVn1h

### DIFF
--- a/packages/obsidian-plugin/tests/e2e/eka-gui/parked-link-placeholder.spec.ts
+++ b/packages/obsidian-plugin/tests/e2e/eka-gui/parked-link-placeholder.spec.ts
@@ -73,10 +73,17 @@ import {
  * plus 5. a RESOLVED link, which must never be checked at all — the placeholder
  * is only reachable from the `if (!file)` seam.
  *
- * Ordering note for the negatives: every link on the page awaits the SAME
- * `ensureIndex()` walk; a miss then returns on a map lookup, while the hit does
- * one further `read()`. So once the placeholder is on the page, the misses have
- * already resolved — polling for the hit is sufficient to read all five.
+ * ⛔ Ordering note — and its LIMIT. Every link on the page awaits the SAME
+ * `ensureIndex()` walk. Controls 2 and 4 then return on a map MISS, so they
+ * settle before the hit. Control 3 does NOT: the non-asset file IS indexed (it
+ * sits two segments under the parked root, so `assetSpaceOf` succeeds) and is
+ * rejected only AFTER its own `await read()` — i.e. at the same async depth as
+ * the hit, with no completion-order guarantee between two concurrent reads.
+ * `data-parked-checked` cannot close that gap either: it is set BEFORE the
+ * lookup, so it proves the decorator was REACHED, not that it finished.
+ * Therefore the snapshot is taken only once the DOM has stopped changing, which
+ * makes "all five verdicts read from one settled render" true by construction
+ * rather than by argument.
  *
  * Render-only: nothing is applied, no profile is switched, and the ephemeral
  * vault is deleted in `afterAll`.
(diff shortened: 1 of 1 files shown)


kitelev:
## Revert-verify — done in CI, on a throwaway branch

Local Docker/QEMU e2e is forbidden on this machine (kernel-panic risk), so the flip was run in CI. I put the break on a **separate branch** (`alp-2858b33e-revertverify` = this PR's head + the break) and dispatched `eka-gui-e2e.yml` against it, so **this PR's branch was never in a state where merging it would ship a broken production file**.

**Break:** the `decorate(...)` call removed from `BodyLinkPatch.patchLink`'s `if (!file)` seam (type-preserving `void`).

| run | body scenario | frontmatter scenario | result |
|---|---|---|---|
| [31742904421](https://github.com/kitelev/exocortex/actions/runs/31742904421) — **with** the break | ✘ RED | ✓ GREEN | 1 failed, 10 passed |
| [PR head `da76f25b`](https://github.com/kitelev/exocortex/actions/runs/31742855878) — **without** | ✓ GREEN | ✓ GREEN | 11 passed |

That asymmetry is the requirement's own Gherkin clause — *«removing the placeholder call from either channel leaves the OTHER channel still working»* — so the two channels are proven independently reachable, not jointly.

### The red run names its own mechanism

The on-timeout diagnostic added in `02a1d965` made one run decisive:

```
TIMEOUT "parked placeholder rendered in the note body" — selector matched 5:
  b1…      text="e2e0da4c-0000-4000-a000-0000000000b1"  checked=null
  …dead    text="e2e0da4c-0000-4000-a000-00000000dead"  checked=null
  not-an-asset                                            checked=null
  dotfolder-decoy                                         checked=null
  seed-area  text="🗂 EKA E2E Seed Area"                  checked=null
```

`checked=null` on every unresolved link is the decorator never being reached, and the parked link falling back to its raw UID is the user-visible cost the feature exists to remove.

### Canary — the dot-folder control is not tautological

The green run logs the adapter's own listing:

```
parked mount listing: {"folders":[".exocortex/parked/kitelev/exoas-parked-e2e/.git",
                                 ".exocortex/parked/kitelev/exoas-parked-e2e/parked"],
                      "files":[".exocortex/parked/kitelev/exoas-parked-e2e/e2e-parked-not-an-asset.md"]}
```

So `vault.adapter.list()` genuinely surfaces `.git` — the walk's skip-guard is live code, and «the decoy stays broken» means something.

### Measured shape of the frontmatter element

```
className: "metadata-link-inner internal-link is-unresolved exocortex-parked-link"
```

Not an `<a>` — which is why both selectors are tag-agnostic, mirroring `BodyLinkPatch`/`PropertiesLinkPatch`'s own.

PR #4047, opened 2026-08-14
fix(grounding): service_call honours targetsCreatedInstance (#4046)

### LINKED ISSUE #4046: service_call ignores targetsCreatedInstance — composite cannot act on the created instance (archive composite leaves file unmoved)
## Symptom

A `composite` grounding that does `[create_instance, …, service_call]` **cannot make the `service_call` act on the created instance**. The step reports success and silently does nothing to the new asset — it operates on the click-target instead.

Concretely: the vault composite `27a37448` ("Create Action instance, execute and archive") creates an `ems__Action`, marks it `archived: true` and re-anchors its `exo__Asset_isDefinedBy` to the archive ontology — but the file **stays physically in the source folder**, breaking the co-location invariant. Adding a 5th step `service_call repairFolder` with `exocmd__Grounding_targetsCreatedInstance: true` (mirroring the canonical `ee48c626` "Archive Ontologically") **does not fix it**: verified on an isolated copy — command returns success, file unmoved.

Failure mode is **quieter than a crash**: no error, success message, invariant broken.

## Root cause (read on `origin/main`)

`targetsCreatedInstance` re-points **`stepPath`** only, and `service_call` never consumes it:

1. `packages/core/src/services/GroundingExecutor.ts` — `executeComposite`:
   ```
   const stepUsesCreatedPath = step.type === BODY_TEMPLATE || step.targetsCreatedInstance === true;
   const stepPath = stepUsesCreatedPath && lastCreatedPath ? lastCreatedPath : filePath;
   ```
   with an explicit comment: *"`targetIRI` is **intentionally NOT re-pointed**: `$target` still resolves to the source asset, and the just-created asset is not yet in the store."*

2. Same file — `executeServiceCall(grounding, targetIRI, filePath, userInput)` dispatches:
   ```
   await service.execute(targetIRI, mergedInput);
   ```
   — **`filePath` is never passed.**

3. `packages/services/src/grounding-service-factories.ts` — `createRepairFolderService`:
   ```
   async execute(targetIRI: string): Promise<void> {
     const targetFile = resolver.resolveFile(targetIRI);   // ← resolves from targetIRI
   ```

⇒ The flag changes a value this path ignores. The step "repairs" the click-target's folder (already correct) and returns success.

This is not specific to `repairFolder` — **every** registered service is affected, because `IGroundingService.execute` has no file-path channel at all.

## Impact

- Any composite wanting `create_instance` → `service_call` on the new asset is impossible to express in vault data.
- The archive-composite path silently produces co-location violations. Measured: 185/185 assets anchored to `$my-efforts/archived` are currently physically in `archived/` — the invariant is clean today, so the first successful run of this composite creates the first violation.
- `audit co-location` would flag it; SHACL does not.

## Proposed fix (one of)

- **A.** Extend `IGroundingService.execute(targetIRI, userInput, filePath?)` and have `executeServiceCall` forward `filePath`; path-based services prefer `filePath` when present. Backwards-compatible (optional param).
- **B.** In `executeComposite`, when `targetsCreatedInstance` is set, resolve the created asset's IRI and re-point `targetIRI` for that step only. Riskier — the comment above says the new asset is not yet in the store, so the IRI may not resolve.

A is the smaller blast radius.

## Acceptance

- [ ] Composite `[create_instance, service_call(repairFolder, targetsCreatedInstance)]` relocates the **created** asset, not the click-target
- [ ] **Revert-verify:** without the fix the file stays in the source folder (RED); with it, it lands in the archive folder (GREEN)
- [ ] Existing `service_call` groundings **without** the flag are byte-identically unaffected (`ee48c626` Archive Ontologically keeps working on the click-target)
- [ ] Negative control: a `service_call` step with the flag but **no** prior `create_instance` still targets the click-target

## Notes

Discovered while shipping one-click "Create, Execute & Archive" (vault-data only, no engine change). The precondition/one-click part is done and released; this is the remaining half — the composite archives **metadata only**, without moving the file.

Workaround until fixed: run "Repair Folder" on the created asset (or `apply repair-folder <path>`).

Vault mirror: `ems__Bug` `2d66698c-d311-448f-bf74-661b43cb6efa` (vault-exodev) carries the full investigation.

kitelev:
Closes #4046.

## Symptom

A `service_call` composite step that opted into `exocmd__Grounding_targetsCreatedInstance` silently operated on the composite **click-target** instead of the asset a previous `create_instance` step produced — and reported **success**. Concretely: the vault composite "Create Action instance, execute and archive" re-anchored the new asset's `exo__Asset_isDefinedBy` but never relocated the file, breaking the co-location invariant with no error and no notice.

## Root cause

`executeComposite` re-points only `stepPath`. Every registered service resolves its file from the IRI it is handed — `IGroundingService.execute` has no file-path channel — so `executeServiceCall` dispatching `service.execute(targetIRI, …)` made the flag a **no-op for this grounding type**. Not `repairFolder`-specific: every service is affected.

## Fix (one site)

When — and only when — the step carries the flag, `executeServiceCall` hands the service `vaultPathToIRI(filePath)`.

That dialect is what **both** target resolvers already accept:
- CLI `createPathBasedTargetResolver` strips it via `iriToVaultPath`;
- plugin `createObsidianTargetResolver` decodes it and looks the path up in `app.vault` (which, unlike `metadataCache`, registers a newly created `TFile` synchronously).

⛔ A bare vault-relative path would **not** work — the plugin resolver treats a non-`obsidian://` input as a uid/`@id` and falls into a metadataCache scan that cannot match a file path.

With the flag absent this is `targetIRI` byte-for-byte, so every existing grounding is unaffected. `$target` substitution inside `serviceCallPayload` deliberately still uses the click-target IRI (link-back semantics, unchanged). The `targetsCreatedInstance` docstring is updated to state this extra hop.

## Tests

New suite in **`packages/cli`** (CI-gated; `packages/services/tests` is not run by CI) driving the REAL `GroundingExecutor`, REAL `createRepairFolderService`, REAL `createPathBasedTargetResolver`:

1. flagged step relocates the **created** asset; click-target untouched
2. no flag → service still receives the click-target path (no regression)
3. negative control: flag set, **no** prior `create_instance` → click-target
4. the handed IRI resolves back to the created file via the production resolver

**Revert-verify:** dropping the branch turns axes **1 and 4 RED** while **2 and 3 stay GREEN** — verified by mutating a copy of the file and restoring it (the working tree was never left in the broken state).

## Local gates

- `packages/core`: **363/363 suites, 8513 tests** green (submodules initialised).
- `packages/cli`: 156/157 suites. The single failure, `sparql-exo003.integration.test.ts`, **reproduces on `origin/main`** — pre-existing, unrelated.
- All three `lint`-job guard scripts pass (shard-assignments, test-antipatterns ratchet, coverage-monotonic).
- `prettier`: the new test file is formatted; `GroundingExecutor.ts` was left alone deliberately — it is **pre-existing prettier-dirty on `origin/main`** (393-line blast radius), and my added lines are already prettier-clean.

⚠ Touches `packages/core/src/services/GroundingExecutor.ts` → per the repo's auto-merge discipline this needs a `code-reviewer` pass before merge; **not** using `--auto`.

https://claude.ai/code/session_01Xp6ceD53nG2HvWFtFWMoFX

### DIFF
--- a/packages/cli/tests/integration/servicecall-targets-created-instance.integration.test.ts
+++ b/packages/cli/tests/integration/servicecall-targets-created-instance.integration.test.ts
@@ -19,11 +19,17 @@
  * axis; the rest stay GREEN, which is what makes them non-vacuous):
  *   - drop the `targetsCreatedInstance` branch in `executeServiceCall`
  *     (back to `service.execute(targetIRI, …)`) → axes 1 and 4 RED;
- *   - drop the `targetQuery` × `targetsCreatedInstance` guard → axis 5 RED;
- *   - re-point `ObsidianFileSystemAdapter.createFile` at `vault.adapter.write`
- *     → the plugin-package unit axis ("must NOT use the raw adapter.write") RED.
- *     Axis 6 here characterises that runtime rather than pinning it: it models
- *     the pre-fix writer and shows the two semantics are distinguishable.
+ *   - drop the `targetQuery` × `targetsCreatedInstance` guard → axis 5 RED.
+ *
+ * The fake below registers a created file in `getAbstractFileByPath`
+ * immediately, which is faithful to BOTH runtimes: the CLI has no index at all,
+ * and in the plugin `FileSystemAdapter.write` awaits `reconcileInternalFile` in
+ * a `finally` → `Vault.onChange("file-created")` → `fileMap[path]`, so the
+ * `TFile` is registered before the write resolves (measured in Obsidian 1.13.7
+ * `app.js`; `Vault.create` adds no registration of its own — it delegates to
+ * that same `adapter.write`). What DOES lag in the plugin is `metadataCache`,
+ * and `createRepairFolderService` already reads frontmatter fresh from disk for
+ * exactly that reason (req 8efc003c) — so this fake models the real split.
  */
 import { describe, it, expect } from "@jest/globals";
 import {
@@ -306,7 +312,7 @@ describe("service_call honours targetsCreatedInstance (Issue #4046)", () => {
 
     const result = await executor.execute(
       composite([createStep(), repairStep(true)]),
-      TARGET_PATH,
+      TARGET_IRI,
       TARGET_PATH,
     );
     expect(result.success).toBe(true);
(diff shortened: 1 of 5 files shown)


kitelev:
## Review round 2 — `d100e461`

Round 1 returned **BLOCK** on a CRITICAL I had missed, and it was right. Verified first-hand before acting on it:

- `ObsidianFileSystemAdapter.createFile` used `vault.adapter.write` — the raw `DataAdapter`, which registers **no `TFile`**;
- `ExocortexPlugin.ts:700,716-718` wires that very adapter as the executor's reader **and** writer;
- so my round-1 IRI (`obsidian://vault/<path>`) hit `createObsidianTargetResolver` → `app.vault.getAbstractFileByPath` → `null` → uid scan → **throw**, and `executeComposite` rolls a failed step back by **deleting `createdPaths`**.

Net effect of round 1 alone in the plugin: "asset created but not relocated" (recoverable via **Repair Folder**) would have become **"error notice and no asset"**, intermittently — strictly worse than the bug.

⛔ And deeper than the review stated: `ObsidianVaultAdapter.rename` → `toObsidianFile` → `vault.getAbstractFileByPath` throws for an unregistered path too. So **variant A from the issue** (a file-path channel on `IGroundingService`) would not have worked either — the created file must be in the Vault index for *any* plugin-side Vault-API operation. I've corrected that in the issue body's proposal list.

### Root fix

`ObsidianFileSystemAdapter.createFile` now creates through **`Vault.create`**, which returns the registered `TFile` and closes the window synchronously. `updateFile` deliberately stays on `adapter.write` (the path is already registered). The deliberate `adapter.write` calls in ExoSync (`SyncCommands`, `VaultRDFIndexer`) are untouched — those bypass the vault event *on purpose*.

### Also addressed from round 1

| Sev | Finding | Action |
|---|---|---|
| CRITICAL | plugin created file invisible to Vault index | root fix above |
| HIGH | comment claimed `app.vault` "registers a newly created TFile synchronously" — false of the writer actually held | reworded; both the comment and the `targetsCreatedInstance` docstring now state that this **depends on** the adapter change, so the halves can't be split silently |
| HIGH | suite modelled CLI semantics only | new plugin-package axis is the gating one; new CLI axis 6 characterises the pre-fix writer |
| MEDIUM | `targetQuery` × `targetsCreatedInstance` unguarded for `service_call` | guard added, mirroring `executePropertySet` |
| LOW | axes passed a bare path where production passes an IRI; private-field reach | both fixed (`vaultPathToIRI`, registry returned from `setup()`) |

### Revert-verify, each mutation in isolation on a copy

| Mutation | Result |
|---|---|
| adapter → `vault.adapter.write` | plugin axis *"must NOT use the raw adapter.write"* **RED** (2 failed / 38 passed); rest GREEN |
| drop the `targetQuery` guard | CLI axis 5 **RED**; other five GREEN |
| drop the IRI branch (round 1 fix) | CLI axes 1 and 4 **RED**; 2 and 3 GREEN |

### Local gates

- plugin **340/340** suites, 6074 tests
- core **363/363**, 8513 tests
- CLI 156/157 — `sparql-exo003.integration.test.ts` reproduces on `origin/main` (pre-existing)
- root `tsc --noEmit` clean; all three `lint`-job guard scripts pass
- prettier: new test formatted; `GroundingExecutor.ts` / both `ObsidianFileSystemAdapter` files left alone — pre-existing prettier-dirty on `origin/main` (393 / 13 / 25 lines), and prettier has no objection to my own lines

Round-2 review is running now. Still **not** `--auto`-merging.

https://claude.ai/code/session_01Xp6ceD53nG2HvWFtFWMoFX

PR #4067, opened 2026-08-16
feat(req-audit): corpus floor — every declared reqs assetspace must contribute

### LINKED ISSUE #4066: req-audit: the active-gate floor measures the PRESENCE of a corpus, not its SIZE — losing a whole assetspace still passes green
## Problem

PR #4065 (req `99e06488-2da6-48fe-bd64-30c1e20bca0f`) gave the blocking `Active-requirement gate` a population floor: `activeTotal === 0` is now an INPUT-integrity failure rather than a vacuous pass. That closes the **total** loss of the corpus.

It does **not** close **partial** loss — and partial loss is the likelier shape. The floor measures the **presence** of a population, not its **size**, so losing an entire assetspace still passes green and silent.

## Measured (execution, on the PR head)

Corpus = the two clones the CI job makes, with `exoas-ems-reqs` replaced by an empty directory (which is exactly what `git clone` of an emptied/renamed repo leaves behind — **the clone step does not fail**):

| corpus | requirementCount | activeTotal | `inputIntegrityViolation` | blocking gate |
|---|---|---|---|---|
| both assetspaces | 172 | 161 | `null` | exit 0 — correct |
| **`exoas-ems-reqs` missing entirely** | **163** | **154** | `null` | ⛔ **exit 0** — `✅ Active-gate OK — 154 active requirement(s), all bound.` |

Nine requirements — a whole module's worth — vanished from enforcement with no signal anywhere.

## Second path to the same silence — malformed YAML is dropped, not raised

`packages/req-audit/src/assets.ts`:

```ts
try {
  metadata = extractFrontmatter(await readFile(abs, "utf-8"), rel);
} catch {
  continue;          // ← the asset silently leaves the corpus
}
```

and `extractFrontmatter` itself returns `null` for genuinely malformed YAML ("Preserve the historical throw→null/`{}` fallback rather than crash the caller"). So a corrupted frontmatter — a bad migration, a stray `: ` in an unquoted scalar — shrinks the corpus by exactly the affected files, silently. The floor only notices if the count reaches **zero**.

## Why this is the boundary of the invariant, not a defect of #4065

The requirement behind #4065 is scoped to zero (`activeTotal === 0`), and it delivers that. The invariant it serves — *"the corpus actually arrived"* — is therefore **half closed**: total loss is loud, partial loss is silent. Filing the other half rather than widening a merged req.

## Candidate fixes (not yet decided)

1. **Per-directory floor** — require that *every* top-level directory under `--reqs` yields ≥ 1 requirement. Directly models "each cloned assetspace arrived", needs no stored state, and fails loudly naming the empty one. Needs a decision on what an intentionally-empty new assetspace should do.
2. **Ratchet on the count** — `activeTotal < committed baseline ⇒ fail`. Strictly stronger (catches partial loss *within* an assetspace), but introduces a number that must be bumped on every legitimate `Deprecated` flip — i.e. its own false-red + maintenance mode. This is the same trade-off #4065 weighed and rejected for the zero case; it may still be right here.
3. **Fail loud on a parse error** instead of `continue` — narrower, orthogonal, and arguably correct regardless: a requirement asset that does not parse is not "absent", it is broken.

(1) and (3) look complementary and stateless; (2) is the only one that catches shrinkage inside a surviving assetspace.

## Acceptance

- a corpus missing one of the cloned assetspaces reds the blocking gate, naming which one
- a requirement asset with malformed frontmatter is reported rather than silently dropped
- the live corpus stays green, byte-identical
- revert-verified on both halves (detection + the gate's consumption of it), as in #4065

Follow-up to #4065. Discovered by adversarial review of that PR.

https://claude.ai/code/session_01FmtVN7YufSHQzKX52jqdva

kitelev:
Closes #4066. Req `bba7bd2b-de7a-4043-bfb8-df56bd3d2a0a` (approved) — the second half of the "the corpus actually arrived" invariant that req `99e06488` (PR #4065) left open.

## The gap

`99e06488` refuses a **total** corpus loss (`activeTotal === 0`). A **partial** loss stays silent and is the likelier one: `git clone` of an emptied or renamed repo exits 0, so the clone step does not fail, and the surviving assetspaces still produce a plausible, fully-bound active count.

Measured by execution on the real clones (the audit + the gate script extracted from the shipped `ci.yml`):

| code | corpus | reqCount | activeTotal | integrity | blocking gate |
|---|---|---|---|---|---|
| `origin/main` | without `exoas-ems-reqs` | 164 | 155 | `null` | ⛔ **exit 0** — "✅ Active-gate OK — 155 active requirement(s), all bound." |
| this PR | without `exoas-ems-reqs` | 164 | 155 | set | ✅ **exit 1**, names `exoas-ems-reqs` |
| this PR | full corpus | 173 | 162 | `null` | ✅ exit 0, unchanged OK line |

Nine requirements — a whole module — left the gate's scope without a single signal.

## What changed

- **`DECLARED_REQS_ASSETSPACES`** — the expected corpus, **declared in code** rather than read off the directories that happen to be on disk. This is the load-bearing design point: a population enumerated from the observed input cannot notice that one of its own members went missing, and the measured failure leaves no directory to enumerate at all. (The literal "each top-level dir under `--reqs` must yield ≥1", as sketched in the issue, is vacuous for exactly this scenario.)
- A declared assetspace contributing 0 requirements now sets `inputIntegrityViolation`, naming the offender and both of its roots. **`ci.yml` is unchanged** — the shipped blocking step already refuses a non-null value, and its "report predates the floor → fail closed" guard still applies.
- Assets that cannot be read, or whose frontmatter fence does not parse, are reported as **dropped** instead of being swallowed by `catch { continue; }` (the second road to the same silence, called out in the issue). A file with **no** fence is still an ordinary document, not a drop. Measured: 0 offenders in the live corpus today, so this is a no-op on current data and a guard for the future.
- The report carries **`corpusByAssetspace`** — the INPUT size next to the finding count — including a declared assetspace that contributed zero.
- `--expect a,b` overrides the declared set for a different corpus.

## Rejected: ratchet-on-count

`activeTotal < stored baseline ⇒ refuse` is the only candidate that also catches shrinkage **inside** a surviving assetspace, but it stores a number that must be bumped on every legitimate `Deprecated` flip — its own source of false red, the trade-off `99e06488` already weighed and declined. That residual gap is named explicitly in the requirement's Non-goals rather than left implied.

## Revert-verified

Mutating a **copy** of the tree, one axis per process; control run first (108/108 green, so the suite genuinely runs):

| mutation | RED |
|---|---|
| declared floor neutralised | 2 (the negative axis + the naming axis) |
| dropped → violation neutralised | 1 |
| loader drop-reporting neutralised | 1 |
| declared list altered | 1 (the `ci.yml` drift guard) |
| per-assetspace seed removed | 1 (the input-size report) |

Every other test — including the 8 pre-existing `@req:99e06488` ones — stays **GREEN** under each mutation: non-vacuity and no-regression in the same run. The back-compat guarantee (a 2-arg `auditTraceability` call does not apply the floor) is what keeps those 8 green.

## Notes

- The gate driver (`extractGateScript` / `runBlockingGate`) moved to `tests/helpers/blocking-gate.ts` and is now shared by both suites instead of duplicated — it still extracts the **shipped** workflow, never a hand-kept copy.
- `requirements-audit.test.ts`'s report helper gains `inputIntegrityViolation`: a **pre-existing** gap on `origin/main` (the field landed with `99e06488` and was never added there), surfaced — not introduced — by the two fields this change adds. Verified against `origin/main`: it fails `tsc -p packages/req-audit` there too.
- ⚠ **`code-reviewer` was structurally unavailable to this session** (the child prompt forbids sub-agents). This PR reads user data (frontmatter/YAML), which is exactly the class where an independent reviewer earns its keep — recommend the orchestrator run it against commit `5ea4bb5c` before merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -304,6 +304,12 @@ jobs:
       # M3-closure (see the block above). `soft` = warn only; `hard` = also
       # block when the P0 checklist is not ramp-ready.
       REQ_GATE: soft
+      # The reqs assetspaces this job's corpus consists of — the SINGLE
+      # enumeration, driving BOTH the clone loop below AND the audit's corpus
+      # floor (`--expect`). Adding a module = one name here, nowhere else: the
+      # set the floor requires is definitionally the set CI clones, so the two
+      # cannot drift apart (issue #4066).
+      REQS_ASSETSPACES: exoas-exo-reqs,exoas-ems-reqs
     steps:
       - name: Checkout
         uses: actions/checkout@v6
@@ -322,15 +328,16 @@ jobs:
 
       - name: Clone reqs assetspaces (public)
         # The audit walks `$RUNNER_TEMP/reqs` recursively, so each per-module
-        # reqs assetspace (RFC 0003 §3.2) is just one more clone line here.
+        # reqs assetspace (RFC 0003 §3.2) is one more name in REQS_ASSETSPACES —
+        # ⛔ do NOT add a clone line here, and do NOT repeat the name anywhere
+        # else in this job: that env var is what the corpus floor requires.
         run: |
           mkdir -p "$RUNNER_TEMP/reqs"
-          git clone --depth 1 \
-            https://github.com/kitelev/exoas-exo-reqs.git \
-            "$RUNNER_TEMP/reqs/exoas-exo-reqs"
-          git clone --depth 1 \
-            https://github.com/kitelev/exoas-ems-reqs.git \
-            "$RUNNER_TEMP/reqs/exoas-ems-reqs"
+          for name in ${REQS_ASSETSPACES//,/ }; do
+            git clone --depth 1 \
+              "https://github.com/kitelev/$name.git" \
+              "$RUNNER_TEMP/reqs/$name"
+          done
 
       - name: Run traceability audit (soft — never blocks until the P3 flip)
         id: audit
@@ -346,11 +353,13 @@ jobs:
           rc=0
           npx tsx packages/req-audit/src/cli.ts \
             --reqs "$RUNNER_TEMP/reqs" --tests . --gate "$REQ_GATE" --output json \
+            --expect "$REQS_ASSETSPACES" \
             > "$RUNNER_TEMP/req-report.json" || rc=$?
           cat "$RUNNER_TEMP/req-report.json"
           echo "--- human-readable (gate=$REQ_GATE) ---"
           npx tsx packages/req-audit/src/cli.ts \
-            --reqs "$RUNNER_TEMP/reqs" --tests . --gate "$REQ_GATE" --output text || true
+            --reqs "$RUNNER_TEMP/reqs" --tests . --gate "$REQ_GATE" --output text \
+            --expect "$REQS_ASSETSPACES" || true
           exit $rc
 
       - name: Active-requirement gate (BLOCKING — SDD feature-lifecycle invariant)
--- a/packages/req-audit/src/assets.ts
+++ b/packages/req-audit/src/assets.ts
@@ -18,6 +18,30 @@ export interface MarkdownAsset {
 }
 
 /**
+ * A markdown file the walk could NOT classify — it was found on disk but its
+ * frontmatter never reached the audit.
+ *
+ * A requirement whose frontmatter cannot be read or parsed is **broken**, not
+ * absent; skipping it silently makes it indistinguishable from a requirement
+ * that never existed (issue #4066, second road to the same silence). Reporting
+ * the drop is what lets the caller refuse instead of guess.
+ */
+export interface DroppedAsset {
+  path: string;
+  /** Why the asset never made it into the corpus (human-readable, printed verbatim). */
+  reason: string;
+}
+
+/** Result of one disk pass: the classified assets plus the ones that were dropped. */
+export interface MarkdownScan {
+  assets: MarkdownAsset[];
+  dropped: DroppedAsset[];
+}
+
+/** The `---`-fenced frontmatter mapping at the very start of a document. */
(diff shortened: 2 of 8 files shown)


kitelev:
**Orchestrator MEDIUM applied** (commit `84e95827`) — single-sourced, with one correction to the record and one finding of my own.

**Correction:** the drift was *not* silent. Adding a third clone line while leaving `DECLARED_REQS_ASSETSPACES` alone reds the drift guard, and that suite runs in the blocking `Test the requirements-audit tool` step of the required `requirements-trace` job — measured by mutating `ci.yml` in a copy, 1 RED. So this change replaces **detection with construction**; it does not close an open hole. Worth doing anyway: by-construction is the shape this whole feature argues for, and leaving the instance one floor up would be inconsistent.

**What changed:** `REQS_ASSETSPACES` joins `REQ_GATE` as a job-level single source (same local precedent). The clone step loops over it; both audit invocations pass it as `--expect`. The set the floor requires is now definitionally the set CI clones. The bash was verified by execution before touching the workflow, and the fail direction is safe — if the loop breaks, nothing clones and the floor fires RED.

**My own finding, from the mutant:** the first version of the guard used `toContain('--expect "$REQS_ASSETSPACES"')`, and dropping `--expect` from the **authoritative JSON** run (whose report the blocking step reads) stayed **GREEN** — the assertion was satisfied by the advisory *text* run. Presence-somewhere, not presence-at-each-site. Now counted: every `cli.ts` invocation in the job must carry the floor.

Revert-verify on the workflow (108/108 unmutated, 1 RED each): re-typed clone line · env≠constant · `--expect` dropped from JSON run (was green before) · `--expect` dropped from text run.

Issue #4267, opened 2026-09-18
perf(cli): triple-cache delta still pays whole-file rewrite, adapter linkpath-index rebuild and a second stat-walk (follow-up to #4263)


kitelev:
## Summary
Follow-up to #4263 / PR #4265 (orchestrator review MEDIUM-3 (3)+(4) and LOW-1 — acknowledged in the PR body, not fixed there). The delta refresh is correct and O(changed) in its own scans, but three fixed costs remain per `--use-cache` command that sees a change, and one on every hit.

## What is still not O(changed)
1. **Whole-cache rewrite on every delta.** `CacheManager.applyDelta` re-serializes and writes the entire `.exocortex/cache/triples.json` (~97 MB on a 16.6 k-file vault; measured 0.14 s stringify + 0.22 s write, plus the 0.5 s read on entry). A per-file segment store (one file per entry, or an append-only journal + compaction) would make a 1-file delta touch ~1 KB.
2. **Adapter linkpath index rebuild.** Any NON-UUID wikilink inside a re-parsed file makes `FileSystemVaultAdapter.buildLinkpathIndex` parse the frontmatter of every vault file (16.6 k YAML parses) to build the basename/alias index — the same cost as a full walk, paid by a delta of one file that happens to contain `[[Some Title]]`. A persisted `basename → path` / `alias → path` index (or deriving it from the cache's own `exo:Asset_aliases` triples + the manifest's paths) removes it.
3. **Two stat-walks on a hit.** `query --use-cache` runs `computeVaultSignature()` (query-result cache, #3983) AND `CacheManager.computeFileManifest()` (triple cache) — two independent walks of the same tree (~70–150 ms each on 16.6 k files). One walk could feed both.

## Numbers (copy of vault-bot-kitelev, 16 652 .md, PR #4265 body)
hit 2.4 s · delta (1 file, layer kept) ≈ 3 s · delta with re-materialization ≈ 5 s · full parse 13.9 s.

## Acceptance
- A 1-file delta writes O(changed) bytes to disk (segment or journal), measured on the same copy.
- A delta of a file with a non-UUID link does not re-parse the whole vault's frontmatter (count `getFrontmatter` calls).
- One tree walk per `query --use-cache` invocation.

## Related
- #4263 (parent), PR #4265, #4264 (write-through consumers).

# code-issue: perf follow-up in packages/cli/src/cache + FileSystemVaultAdapter (не vault-данные)

kitelev:
[...]

Issue #4263, opened 2026-09-18
fix(cli): CacheManager invalidates by vault-root mtime — misses nested asset edits; migrate to file-signature + delta refresh


kitelev:
## Summary
`CacheManager.isCacheValid()` (`packages/cli/src/cache/CacheManager.ts`) invalidates the persistent triple cache by comparing the vault **root directory** mtime only, so it silently misses nested `assetspaces/**` edits and can never support incremental (delta) refresh — fix it by reusing the existing per-file signature primitive (`computeVaultSignature()`, shipped for the query-result cache in #3983) and adding delta refresh built on the existing `NoteToRDFConverter.convertNote()` per-file API.

## User Story / JTBD
**As a** CLI/automation consumer that runs several `exocortex-cli` processes back-to-back against the same vault (e.g. a Telegram-bot orchestration loop calling `apply` → `apply` → `resolve-buttons` in one user turn)
**I want** the persistent triple cache to detect nested file changes correctly and refresh only what changed
**So that** I don't have to run `index --force` (full re-parse, ~12-57s on a 16.6k-file vault) before every command just to be safe, and repeated processes in the same turn pay a near-zero incremental cost instead of a full re-parse each time.

## Background
**Related measurement (2026-09-18, vault-bot-kitelev: 16 651 `.md` files, 275 384 triples, CLI 16.240.15):** a consumer (`~/dotfiles/scripts/olya-assistant/poll_assistant.py`) currently runs `exocortex-cli index --force` before every turn (median 20s, up to 57s) specifically because it cannot trust `CacheManager`'s cache-validity check across nested edits — see the comment in that script's `build_cache()`.

The root cause is verified in `origin/main`:
- `CacheManager.isCacheValid()` (`packages/cli/src/cache/CacheManager.ts`, method starting at the `async isCacheValid()` declaration) does:
  ```ts
  const vaultStats = await fs.stat(this.vaultPath);
  if (vaultStats.mtimeMs !== cacheData.metadata.vaultMtime) { return false; }
  ```
  i.e. it stats **only the vault root directory**, not the files under it.
- This exact blind spot has **already been diagnosed and fixed once, but for a different cache.** `packages/cli/src/cache/vaultSignature.ts` (added by #3983, "feat(cli): vault-mtime-aware invalidation for SPARQL query-result cache") implements `computeVaultSignature()` — a SHA-256 over the sorted `relPath:mtimeMs` list of every `.md` file — and its docstring says explicitly:
[...]

Issue #3788, opened 2026-06-28
CLI: apply status-transitions (start-effort etc.) reject valid Backlog/ToDo tasks — precondition ASK is true in store but apply returns 'Precondition not satisfied'


kitelev:
## Summary

`npx @kitelev/exocortex-cli apply start-effort <task>` (and the other gomoiconic status-transition commands) **rejects a valid Backlog task** with `❌ Precondition not satisfied`, even though the command's own SPARQL-ASK precondition is **logically true** against the freshly-loaded store. The same transition works in the Obsidian plugin. This is a CLI↔plugin parity bug that makes the gomoiconic task lifecycle **undrivable from the CLI** (blocks dogfooding / executive-assistant task tracking).

CLI version observed: **16.166.5**.

## Repro (deterministic)

```bash
# 1. Create a standalone ems__Task WITH an explicit Backlog status
npx -y @kitelev/exocortex-cli@16.166.5 create --vault <vault> \
  --class 1b20a8f0-d745-4e93-91db-4531b3df120e \
  --label "Repro task" \
  --created-by de20a3f1-7483-4714-ab28-b45f5cf02c76 \
  --property "exo__Asset_isDefinedBy=[[<some-onto-uid>]]" \
  --property "ems__Effort_status=[[753a44d5-846c-4b82-9196-4fd9a4d48777]]"   # ems__EffortStatusBacklog

# 2. Try the gomoiconic start transition
rm -f <vault>/.exocortex/cache/triples.json
npx -y @kitelev/exocortex-cli@16.166.5 apply start-effort <task-path> --vault <vault> --yes
# => ❌ Precondition not satisfied for "Start Effort" on "<task-path>".
```

## Evidence that the precondition SHOULD pass

The precondition is asset `575404fc-9085-4f4a-a3e4-bfebb49c42a4` ("Allow transition to Doing (from ToDo / Backlog)"), `exocmd__Precondition_sparqlAsk`:

```sparql
PREFIX ems: <https://exocortex.my/ontology/ems#>
ASK {
  FILTER NOT EXISTS {
    $target <https://exocortex.my/ontology/exo#Instance_class> ?protoMarkerClass .
    FILTER(STRENDS(STR(?protoMarkerClass), "Prototype"))
  }
  $target ems:Effort_status ?s .
  FILTER(?s IN (
    <https://exocortex.my/ontology/ems#EffortStatusToDo>,
    <https://exocortex.my/ontology/ems#EffortStatusBacklog>
[...]

PR #4270, opened 2026-09-19
feat(cli): --use-cache on apply / resolve-buttons / create; write-through opt-in via --write-through (#4264)

### LINKED ISSUE #4264: feat(cli): wire --use-cache to apply/resolve-buttons/create with write-through (blocked by #4263)
## Summary
`apply`, `resolve-buttons`, and `create` each rebuild the whole-vault triple store from scratch on every invocation (verified: no `--use-cache`/`useCache` anywhere in `apply.ts`, `resolve-buttons.ts`, `create.ts` on `origin/main`) — wire the persistent triple cache from #4263 into these three commands via `--use-cache`, add write-through so a mutating `apply`/`create` call keeps the cache warm for the next process in the same turn, and prove parity with a revert-verify suite across the exact multi-process chain a consumer (Telegram-bot orchestration) already runs.

## User Story / JTBD
**As a** CLI/automation consumer that chains several `exocortex-cli apply`/`resolve-buttons`/`create` processes in one logical turn (e.g. `create-task-instance` → `move-to-backlog` → `start-effort` to log one user action)
**I want** each of those processes to reuse a persistent, correctly-invalidated triple cache instead of re-parsing the whole vault every time
**So that** a 3-command chain that today costs ~80s of pure CLI parse time (measured: `create-task-instance` 38.8s median, `move-to-backlog` 58s median, `resolve-buttons --json` 30.4s median on a 16.6k-file / 275k-triple vault) drops to near-idle parse cost (~12-13s baseline) after the first process warms the cache, with correct read-after-write behavior across processes.

## Background
**Measurement (2026-09-18, vault-bot-kitelev: 16 651 `.md` files, 275 384 triples, CLI 16.240.15):** a Telegram-bot engine (`~/dotfiles/scripts/olya-assistant/poll_assistant.py`) processes one user message in a median of 134s. Of the 89s median Claude-turn time, ~70% (1753s CLI vs 709s model, over 25 sampled turns) is spent inside `exocortex-cli` subprocesses, each paying a full vault re-parse:
- `apply create-task-instance` median 38.8s (max 75.6s)
- `apply mark-done` median 19.4s
- `apply move-to-backlog` median 58s
- `apply set-event-time` median 24.4s
- `apply create-execute-archive` median 62.9s
- `resolve-buttons --json` median 30.4s (max 83s)
- On an otherwise-idle machine the same commands cost ~12-13s (baseline full parse); under concurrent load the cost is 3-6x that baseline — i.e. this is a repeated-process-cost problem, not (primarily) a parse-algorithm problem.

Code facts verified on `origin/main`:
- `apply.ts` (comment "Build triple store once for the whole batch"): `new FileSystemVaultAdapter(vaultPath)` → `new NoteToRDFConverter(vaultAdapter)` → `.convertVault()` → `InMemoryTripleStore.addAll()`, unconditionally, on every invocation.
- `resolve-buttons.ts` (comment "Build the triple store from the whole vault (same machinery as `apply`)"): identical unconditional full-parse pattern.
- `create.ts`: same `FileSystemVaultAdapter`/`NoteToRDFConverter` full-parse pattern, plus `ShapeLoader.loadFromVaultFS(vaultPath)` (separate SHACL-shape load, not a cache candidate for this issue — noted as out of scope below).
- None of the three files reference `useCache`/`--use-cache` anywhere; the only existing `--use-cache` consumers are `sparql-query.ts`, `classes.ts`, `run-query.ts`, `validate-schema.ts`.
- A regression precedent for stale-cache-vs-mutation exists in this codebase's own issue history: #3788 ("Precondition not satisfied" from stale state) — the write-through requirement in this issue's AC3 exists specifically to prevent reintroducing that class of bug via a mutation-unaware cache.

## Related Issues
- Part of: standalone perf fix (no parent Project)
- **blockedBy: #4263** — this issue's mutating commands (`apply`, `create`) cannot safely read from a persistent cache whose validity check misses nested edits; #4263 fixes that and also adds the delta-refresh + shared-loader groundwork this issue's write-through builds on.
- Related: #3788 (stale-state precondition failure) — the concrete regression class the write-through requirement (AC3) exists to prevent.
- Related: #4195 (open, `--cached --ttl` for query results) — different cache, different lifecycle (TTL-based query-result cache vs write-through triple cache); not a substitute for this issue.

## BoK References
| Body of Knowledge | Chapter/Section | Relevance |
|--------------------|-----------------|-----------|
| SWEBOK v3 | Ch. 4 Software Testing | Revert-verify regression suite across the exact multi-process mutation chain (`create-task-instance` → `move-to-backlog` → `start-effort`) that motivated this issue |
| SWEBOK v3 | Ch. 2 Software Design | Write-through cache-consistency design for a cache now shared by both read (`resolve-buttons`) and write (`apply`, `create`) paths |
| SWEBOK v3 | Ch. 10 Software Quality | Output parity (byte-identical `--dry-run`/`--json`) between cached and non-cached paths is a correctness, not just performance, requirement |
| ISTQB | Test Techniques | Black-box equivalence testing: cached vs non-cached invocation of the same command on the same vault state must be observationally indistinguishable except for latency |
| PMBOK v7 | Domain: Delivery | Downstream consumer (bot engine) wiring is an explicit follow-up noted in this issue, sequenced after CLI release, not bundled into this PR |

### Architecture Context
Builds directly on #4263's `loadVaultTriples(vaultPath, { useCache })` shared helper and delta-aware `CacheManager`. This issue's scope is: (1) expose `--use-cache` on three more commands, (2) make mutations (`apply`, `create`) keep the cache warm for the *next* process (write-through), (3) prove parity + the specific multi-process read-after-write chain.

### Implementation Steps
1. Add the `--use-cache` CLI option to `apply.ts`, `resolve-buttons.ts`, `create.ts` (mirroring the existing option definition/description on `sparql-query.ts`), and route their triple-store construction through #4263's `loadVaultTriples()` helper instead of the direct `FileSystemVaultAdapter`/`NoteToRDFConverter`/`convertVault()` sequence, guarded by the flag (default off, so today's behavior is unchanged without the flag).
2. Implement write-through: after `apply`/`create` successfully writes a file mutation, update the in-process triple store *and* persist the delta to the on-disk cache (reusing #4263's delta-refresh machinery: drop the mutated file's own-subject triples, re-run `convertNote()` on the just-written file, merge, persist) — so the next `exocortex-cli` process in the chain reads a cache that already reflects this process's own write, without needing to re-walk the whole vault to detect it.
3. `create.ts`'s `ShapeLoader.loadFromVaultFS()` (SHACL shapes) is a **separate load path, out of scope for this issue** — `--use-cache` here only covers the triple-store build, not shape loading. State this explicitly in the PR description to avoid scope creep.
4. Instrument `--use-cache` runs (behind `--stats` or equivalent, following the existing `sparql-query.ts --stats` convention) with a cache-hit/miss + delta-size log line, so the downstream consumer wiring (out of scope for this PR, tracked separately) can verify it's actually hitting the fast path in production.

### Code Example
```typescript
// packages/cli/src/commands/apply.ts (illustrative)
const useCacheEffective = options.useCache ?? false;
const triples = await loadVaultTriples(vaultPath, { useCache: useCacheEffective }); // from #4263
// ... existing store build from `triples` unchanged ...

// after a successful mutation write:
if (useCacheEffective) {
  await cacheManager.refreshFile(mutatedFilePath); // delta refresh for exactly this file, from #4263
}
```

## Techniques Applied
- **Write-through caching** — mutation updates the persisted cache synchronously as part of the command's own write, not lazily on next read.
- **Feature flag / additive rollout** — `--use-cache` defaults off; existing (no-flag) behavior is untouched, matching the pattern already established for `query`/`classes`/`run-query`/`validate-schema`.
- **Black-box equivalence testing** — cached and non-cached runs of the same command/input must produce byte-identical `--dry-run`/`--json` output.
- **Regression chain testing** — the specific 3-process mutation sequence that motivated this issue is captured as a named integration test, not just isolated per-command unit tests.

### Unit Tests
- `--use-cache` flag parsing on each of `apply`, `resolve-buttons`, `create` (option registered, defaults to `false`/unset).
- Write-through: after a mutating `apply`/`create` call with `--use-cache`, the on-disk cache file's manifest/triples reflect the mutation (assert directly on the persisted cache content, not just observed CLI output).

### Integration Tests
- **Parity (per command, `--use-cache` vs without), on a live/fixture vault:** `apply create-task-instance`, `apply move-to-backlog`, `apply start-effort`, `apply mark-done`, `resolve-buttons --json` each produce identical `--dry-run`/`--json` output with and without `--use-cache`.
- **Read-after-write chain (the motivating regression, mirrors #3788's failure class):** run `apply create-task-instance --use-cache` in one process, then `apply move-to-backlog <same-asset> --use-cache` in a **second, separate process**, then `apply start-effort <same-asset> --use-cache` in a **third process** — the second and third processes' precondition checks (`ASK` evaluation) MUST see the first/second process's write, not stale pre-mutation state. This must be a genuine multi-process test (separate `exec`/`spawn` invocations), not an in-process call sequence, to actually exercise the persisted-cache read path.
- **Revert-verify (mandatory):** on pre-fix code (no write-through), the read-after-write chain test above MUST fail (stale precondition) — RED half. Post-fix, it MUST pass — GREEN half. Both committed as the permanent regression test.

### BDD Scenarios
```gherkin
Feature: --use-cache on apply/resolve-buttons/create with write-through consistency

  Scenario: Cached and non-cached runs produce identical output
    Given a vault in a known state
    When I run "apply move-to-backlog <asset> --dry-run --json" with --use-cache
    And I run the same command without --use-cache
    Then both outputs are byte-identical

  Scenario: A mutation is visible to the very next process via the persisted cache
    Given a vault with a persistent triple cache warmed by a prior --use-cache run
    When process A runs "apply create-task-instance <task> --use-cache" and writes a new asset
    And process B (a separate process) runs "apply move-to-backlog <that-asset> --use-cache"
    Then process B's precondition check sees the asset process A just created
    And process B does not report "Precondition not satisfied" due to stale cache state

  Scenario: Default behavior is unchanged without the flag
    Given a vault in a known state
    When I run "resolve-buttons --json" without --use-cache
    Then it performs a full vault parse exactly as before this change
```

## Deliverables
- [ ] `--use-cache` option added to `apply.ts`, `resolve-buttons.ts`, `create.ts`
- [ ] All three routed through #4263's `loadVaultTriples()` helper
- [ ] Write-through implemented for `apply`/`create` mutation paths
- [ ] `--stats`-style cache-hit/miss/delta-size log line for production verification
- [ ] Parity tests (cached vs non-cached) for `create-task-instance`, `move-to-backlog`, `start-effort`, `mark-done`, `resolve-buttons`
- [ ] Multi-process read-after-write chain integration test, with revert-verify RED/GREEN pair
- [ ] `create.ts` scope note (shape loading excluded) documented in code comment and PR description
- [ ] Release notes entry

## Quality Criteria
- Second and later `--use-cache` processes in a warm-cache chain complete in the same order of magnitude as the idle-machine baseline (~12-13s), not the loaded-machine 38-83s range currently measured.
- Zero output divergence between cached and non-cached paths across all five instrumented commands.
- Zero stale-read regressions in the multi-process chain test (this is the primary correctness bar, not just a performance one).

## Acceptance Criteria
- [ ] `apply`, `resolve-buttons`, `create` each accept `--use-cache` and, when set, build their triple store via #4263's cached/delta-aware loader.
- [ ] `--dry-run`/`--json` output for `create-task-instance`, `move-to-backlog`, `start-effort`, `mark-done`, `resolve-buttons` is byte-identical with and without `--use-cache`, on the same vault state.
- [ ] A three-process chain (`create-task-instance` → `move-to-backlog` → `start-effort`, each a separate `--use-cache` invocation) completes with every precondition check correctly seeing the prior process's write — no stale-state failures.
- [ ] Without `--use-cache`, all three commands' behavior is byte-identical to pre-change (default-off flag).
- [ ] Revert-verify RED/GREEN pair for the read-after-write chain exists and is committed.

## Definition of Done
- [ ] Implementation complete per Technical Approach
- [ ] All Test Plan items passing in CI (`test-coverage`, `typecheck`, `lint`, `archgate`, relevant `e2e-shard`)
- [ ] Multi-process revert-verify pair merged as a permanent regression test
- [ ] `create.ts` shape-loading scope note documented
- [ ] PR merged to main; new CLI version published via standard release automation
- [ ] (Tracked separately, not blocking this issue) consumer wiring follow-up filed for `~/dotfiles/scripts/olya-assistant/poll_assistant.py` / `kitelev-exo` wrapper to actually pass `--use-cache` and drop the pre-turn `index --force`

## RACI
| Activity | Responsible | Accountable | Consulted | Informed |
|----------|-------------|-------------|-----------|----------|
| Implementation | AI Agent | kitelev (maintainer) | - | - |
| Testing (unit + integration + multi-process revert-verify) | AI Agent | kitelev (maintainer) | - | - |
| Code Review | AI Agent (self-review via CI gates) | kitelev (maintainer) | - | - |
| Release | AI Agent (automated) | kitelev (maintainer) | - | - |

## Risks
| Risk | Probability | Impact | Mitigation |
|------|-------------|--------|------------|
| Write-through introduces a race between the writing process persisting its delta and a concurrent second process reading the cache mid-write | Medium | High (torn/corrupt cache read by a concurrent process) | Persist via the existing atomic write pattern already used for `.exocortex/cache/triples.json` (write-temp-then-rename, if not already so — verify and harden in this PR); cover with a concurrent-access test |
| `create.ts`'s SHACL `ShapeLoader.loadFromVaultFS()` gets conflated with triple-store caching, silently going stale | Low | Medium (SHACL validation against outdated shapes) | Explicit scope note (Technical Approach step 3) keeps shape loading untouched by this PR; add a code comment at the call site |
| Multi-process integration test is flaky under CI's shared-runner timing | Medium | Low (CI noise, not a correctness gap) | Use explicit process-exit synchronization (await child process completion) rather than sleep/poll, per repo convention |

## Rollback Plan
1. Revert the PR (self-contained change to 3 command files + write-through logic in the shared cache layer from #4263).
2. `--use-cache` is opt-in and default-off on all three commands — a rollback that only reverts this PR (keeping #4263) leaves those three commands back at unconditional full-parse behavior, with no cache-file schema impact.
3. Confirm via a full-parse-only run (`apply <cmd> <path>` without `--use-cache`) that output is unchanged from pre-PR baseline.

## Dependencies
- **blockedBy**: #4263 — file-signature validity check, delta refresh, and shared `loadVaultTriples()` helper this issue's write-through and `--use-cache` wiring depend on.
- **Enables**: consumer wiring in `~/dotfiles/scripts/olya-assistant/poll_assistant.py` / `kitelev-exo` (tracked separately, outside this repo, not a GitHub Issue here).

## Estimates
| Task | Effort |
|------|--------|
| `--use-cache` flag + `loadVaultTriples()` wiring on 3 commands | 1.5h |
| Write-through implementation (delta persist on mutation) + atomic-write hardening | 2.5h |
| Parity tests (5 commands, cached vs non-cached) | 1.5h |
| Multi-process read-after-write chain test + revert-verify RED/GREEN pair | 2h |
| Review fixes / CI stabilization | 0.5h |
| **Total** | **8h** |

## Labels
`enhancement`, `performance`, `package:cli`, `cli`, `size/M`

## Best Practices Checklist
- [ ] `--use-cache` defaults off (additive rollout, no behavior change without the flag)
- [ ] Write-through avoids reintroducing the stale-precondition class of bug seen in #3788
- [ ] `create.ts` shape-loading path explicitly excluded from scope (no silent conflation of two different caches)
- [ ] Multi-process (not in-process) integration test for the read-after-write chain — the thing that actually matters for the motivating bot use case
- [ ] Revert-verify RED/GREEN pair committed per repo's `integration-test-revert-verify` convention

## Review Checklist
- [ ] Code follows project conventions (matches existing `--use-cache` option style on `sparql-query.ts`)
- [ ] Tests are comprehensive per Test Plan, including the mandatory multi-process revert-verify pair
- [ ] No behavior change to any of the three commands when `--use-cache` is not passed
- [ ] `create.ts` scope note present in code and PR description
- [ ] Ready for merge — all required CI checks green

kitelev:
Closes #4264. Requirement: `cb707868-356f-495d-825a-182e66ba8bcd` (kitelev/exoas-exo-reqs, **Approved** → Active after merge).

## What

`--use-cache` on `apply`, `resolve-inline-buttons` (alias `resolve-buttons`) and `create` (`--validate` load + the write path): the triple store comes from `<vault>/.exocortex/cache/triples.json` through the shared `loadVaultTriples()` loader — hit / delta / rebuild — instead of an unconditional full vault parse. Default off; without the flag the three commands are byte-identical to before (no cache read, no cache write).

Two modes for a **mutating** `apply` / `create` under the flag (decision **ae0b4fce**, taken by measurement — matrix below):

| invocation | writer | next `--use-cache` process |
|---|---|---|
| `--use-cache` (**default: delta-only**) | never touches the cache file | pays the delta itself (changed files + referrers re-parsed), its preconditions see the write |
| `--use-cache --write-through` | folds the mutation into the cache right after the grounding executed (`CacheManager.refreshAfterWrite`: manifest diff vs the loaded state → same `planDelta`/`applyDelta` → atomic `tmp+rename`) | plain hit |
| `--write-through` alone | refused before anything is read or written: exit 2, one stderr line | — |

Either way the #3788 class (a precondition ASK evaluated on a store that predates the previous command's write) is closed: the next process sees the write through a hit or through a delta. Write-through is best-effort by construction (`writeThroughCache` never throws; rc and stdout do not depend on it); a rebuild-class change is left to the next reader; an absent cache is never built by a mutating command; a concurrent writer's fresher cache is folded into, never reverted (stamp guard on the cache file's `{mtimeMs,size}`). One stderr line per cache phase; stdout untouched. Documented divergence: on an `index`-built cache the flagged store additionally carries the inferred layer, as `query --use-cache` already does (`index --no-inference` gives the explicit graph).

`applyDelta` now deserializes the merged explicit triples lazily and a write-through asks for none of them unless the inferred layer must be re-materialized (the command keeps its own store) — measured −0.3 s / −50 MB per write-through without inference, no change with inference (the recompute still needs the merged set).

## Why delta-only is the default (measurement)

Setup: `cp -a` copy of `vault-bot-kitelev` (16 654 files; never the live vault), the CLI bundle built from this branch's head (`packages/cli/dist/index.js`, 1.6 MB), the bot's own chain `apply create-task-instance <prototype> --json --yes` → `apply move-to-backlog <created> --json --yes` → `apply start-effort <created> --json --yes` → `resolve-buttons <created> --json`, **each step a separate `node` process**, wall = whole process via `/usr/bin/time -l` (node start-up included), RSS = `maximum resident set size` of that process. Each scenario starts from a fresh copy with no cache and warms it with the bot's `index --force` (with and without `--no-inference`). Script: `measure-4264.py` (in the session scratchpad; the two earlier rounds on the pre-optimization bundle gave the same picture ±0.3 s). Runner: this Mac, nothing else running.

| scenario (each step a separate process) | create-task-instance | move-to-backlog | start-effort | resolve-buttons --json | next pure reader (`query --use-cache`) |
|---|---|---|---|---|---|
| no flag (full parse per process) | 33.0 s / 1036 MB | 11.4 s / 797 MB | 11.8 s / 805 MB | 11.6 s / 846 MB | — |
| `--use-cache` delta-only, cache from `index --no-inference` | 21.6 s / 1313 MB hit | 2.9 s / 1638 MB delta | 2.8 s / 1637 MB delta | 3.1 s / 1664 MB delta | 2.3 s / 1206 MB |
| `--use-cache --write-through`, cache from `index --no-inference` | 22.7 s / 1881 MB hit +WT | 3.2 s / 1846 MB hit +WT | 3.2 s / 1846 MB hit +WT | 2.5 s / 1167 MB hit | 2.3 s / 1206 MB |
| `--use-cache` delta-only, cache from `index` (inference, the bot's current warm-up) | 22.2 s / 1359 MB hit | 5.6 s / 1869 MB delta | 5.2 s / 2081 MB delta | 5.5 s / 2082 MB delta | 2.5 s / 1303 MB |
| `--use-cache --write-through`, cache from `index` (inference) | 26.2 s / 2440 MB hit +WT | 5.7 s / 2716 MB hit +WT | 6.1 s / 2418 MB hit +WT | 2.8 s / 1260 MB hit | 2.6 s / 1303 MB |

Warm-up paid once per bot turn today: `index --force --no-inference` 13.0 s / 1224 MB; `index --force` (inference) 16.7 s / 2431 MB. Cache: 16 654 files, 90 MB (no inference) / 97 MB (21 159 inferred triples).

Reading it:

- One delta per mutation is paid **either way** — the flag only decides *who* pays it. Delta-only: the next `--use-cache` process pays it on load (the `delta` cells). Write-through: the writer pays it after its mutation (`+WT`), and every later reader is a `hit`.
- **On the bot's chain (3 writers in a row) delta-only is cheaper and lighter:** 4-process totals 30.4 s vs 31.6 s (no inference) and 38.5 s vs 40.8 s (inference); writer peak RSS ≤ 1.66 GB vs ≤ 1.88 GB and ≤ 2.08 GB vs ≤ 2.72 GB. The extra memory of write-through is structural — with inference on, `refreshAfterWrite` re-materializes the inferred layer in a second store while the command's own store is still alive; with five bots on one host that 0.6 GB per writer is the number that matters.
- **Write-through wins only when readers outnumber writers:** every additional pure reader after a mutation saves 0.6 s (no inference) / 2.7–2.9 s (inference) and 0.4–0.8 GB, because it hits instead of paying the delta. That is a property of the consumer's workload, not of the CLI — hence opt-in (`--write-through`), not default (decision ae0b4fce).
- Against today (no flag, full parse per process, 33 / 11.4 / 11.8 / 11.6 s): either mode takes the three status steps from ~11.5 s to 3–6 s; `create-task-instance` stays ~22 s because its cost is the prototype instantiation, not the store. The bot's per-turn `index --force` (13–17 s) becomes unnecessary under either mode — consumer wiring, follow-up outside this repo.
- `applyDelta`'s lazy deserialization (this PR) took write-through from 3.4–3.7 s to 3.2 s and −50 MB per writer without inference; with inference nothing changes because the recompute needs the merged set anyway.

### N2 — reader peak RSS, published main (16.240.19) vs this branch

Same vault copy, same cache file (warmed by this branch's `index --force`), `/usr/bin/time -l`. `query --use-cache` is the only reader that has the flag on main; `resolve-buttons` / `apply` have no cache path on main (their main reference is the no-flag full parse in the matrix above), so they are shown branch-only.

| reader | cache from `index --no-inference` | cache from `index` (inference) |
|---|---|---|
| main `query --use-cache` hit (2 runs) | 2.3 s / 1206 MB · 2.3 s / 1205 MB | 2.6 s / 1303 MB · 2.6 s / 1303 MB |
| **branch** `query --use-cache` hit (2 runs) | 2.3 s / 1205 MB · 2.3 s / 1206 MB | 2.5 s / 1303 MB · 2.5 s / 1304 MB |
| main `query --use-cache` delta (1 file) | 3.0 s / 1694 MB | 5.4 s / 2081 MB |
| **branch** `query --use-cache` delta (1 file) | 3.2 s / 1698 MB | 5.1 s / 2080 MB |
| branch `resolve-buttons --use-cache` hit / delta | 2.6 s / 1171 MB / 3.2 s / 1660 MB | 2.6 s / 1266 MB / 5.5 s / 2080 MB |
| branch `apply create-task-instance --use-cache` (delta-only writer, load = hit) | 22.5 s / 1246 MB | 22.6 s / 1321 MB |
| branch `query --use-cache` right after that write (pays the delta) | 3.1 s / 1698 MB | 5.2 s / 2082 MB |

Main and branch are identical within ±4 MB on every reader: the process's peak is reached during the load, when the parsed JSON and the materialized `Triple[]` are alive together, and that peak is the same whether or not the snapshot is retained afterwards. No retention change was made (a `keepCacheManager: writeThrough` switch would buy nothing measurable); the reference stays `this.loaded`-retained for every `--use-cache` load so a write-through in the same process diffs against what it loaded.

## Orchestrator review of 1e8e8204 — response

| finding | resolution |
|---|---|
| **HIGH** write-through persisted from `this.loaded` and reverted a concurrent `index` (inferred layer + `inferenceEnabled`) | stamp guard: `readCacheData` stats the cache file **before** reading and retains `{data, stamp}` together; `writeCacheData` re-stamps from the temp file before its rename; `refreshAfterWrite` trusts the snapshot only while the current stat equals the retained stamp, otherwise re-reads once. Axes **A7b** (concurrent `saveInferredTriples`: flag + layer survive, exactly 1 re-parse), **U10** (state and stamp move together through an intermediate `isCacheValid()`), **U11** (no intermediate read: one fallback read). Mutants M1–M4 (retention / stamp dropped or zeroed) → RED. |
| **MEDIUM-1** AC9 — two stderr lines from a mutating command | AC9 rewritten as *one line per cache phase* with the count pinned per command × mode (A9 matrix: apply mutating 2 / dry-run 1 / refused 1; resolve-buttons 1; create --validate real 2 / dry 1 / bare real 1 / bare dry 0 — under `--write-through`; without it the write-through line never appears). |
| **MEDIUM-2** AC2/AC8 byte-identity false on an `index`-built cache | AC2 qualified in the req + axis **A2b**: a command gated on an inherited property is hidden without the flag and visible with it on an index-built cache; a cache the flag built itself has no layer and stays byte-identical; the write-through keeps the layer (re-materialized only when an engine input changed — #4263's gate). Non-goal added: making the two stores identical would make `apply --use-cache` see less than `query --use-cache`. |
| **MEDIUM-3** measurement matrix {write-through on/off × inferenceEnabled} + peak RSS on an `index`-built cache | done — table above (`/usr/bin/time -l`, both modes, both inference configurations, RSS per process); it is what decided the default. |
| **L1** write-through gated on `successCount > 0` | gated on **executed** (`TargetResult.executed`: a dry-run / refused precondition never executes; a failed composite that landed part of its files does) — axis **A3c**, mutant P5 → RED. |
| **L2** `loadVaultTriples({cacheManager})` ignored `vaultPath` | throws when the caller-owned manager belongs to another vault — axis **L3**, mutant L6 → RED. |
| **L3** `create --use-cache` (bare) reads the whole cache for one file; `--validate` load-delta + write-through = two persists in one process | by design for this PR, named here; the cost lives in the cache format / delta path → **#4267**. Delta-only default removes the second persist from the default path entirely. |
| **L4** `create.test.ts` option-count axis is not AC coverage | agreed — it is registration only (now 17 options + `--write-through` registration); AC coverage is the A/U/L/P axes below. |

## Orchestrator review round 2 (@ b8254b80) — response

| finding | resolution |
|---|---|
| **N1** (must-fix) this branch's `fs.stat(tmp)` broke the anchor of M12 (`write in place, not atomic`) in `cache-manifest-delta-4263.spec.json` | M12 re-anchored on the `stat(tmp)` + `rename` pair (mutant writes in place and removes the temp file, keeps the stamp); the whole #4263 spec (32 mutants, shared subject `CacheManager.ts`) run on this branch — result in the driver section. |
| **N2** (must-fix) snapshot retention now on every `--use-cache` reader and `index` — peak RSS vs main not measured | measured (table above): identical within ±4 MB on hit and delta, both inference configurations — the load peak already holds JSON + triples together. No code change. |
| **N4** cache deleted after load (`index --force` → invalidate) → skipped, not locked | **U13** + mutant **M14** (`loadedIsCurrent` trusts the snapshot regardless of the stat) → RED. |
| **N5** bare `create --use-cache` constructed a CacheManager for nothing | constructed only when `--validate` or `--write-through` will use it. |
| **N6a** AC9 pin "create --validate real without --write-through = 1" not asserted | `cv0` added to A9. |
| **N6b** no proc-chain mutant expecting P10 | **PP4** (refusal guard removed) → P10. |
| **N3** (LOW) stat→rename window spans the whole re-parse | accepted as the lock-free window #4263 already has (Known limits); a re-stat before `rename` is a cheap follow-up if it ever bites — not done here. |
| AC1 wording "CacheManager never constructed" for `create` | stronger than what is locked (for apply / resolve-buttons it is locked transitively by P1 / R1) — a documentation nuance, noted here; the req is not re-worded. |

## Guard → axis / mutant (revert-verify)

Driver: `python3 -I ~/.claude/bin/mutant-driver.py packages/cli/tests/integration/use-cache-4264.<subject>.spec.json` (set `tree`/`subject` to your checkout). rc 0 = control 0 red and every mutant reddens what its `expect` names. Final runs on this branch's head (code = c6421edf; the #4264 specs plus the whole #4263 spec, which shares the subject):

| subject | mutant (guard removed / inverted) | expected RED | driver RED | rc |
|---|---|---|---|---|
| `packages/cli/src/cache/CacheManager.ts` (control red: 0) | M1 read does not retain snapshot | U2, U10, A8 | A8, U10, U2 | 0 |
|  | M2 persist does not retain published state | U6, U9, U11 | U11, U6, U8, U9 |  |
|  | M3 stamp guard removed (stale snapshot trusted) | U11, A7, A7b | A7, A7b, U11 |  |
|  | M14 snapshot trusted even after the cache file was deleted | U13 | A7, A7b, U11, U13 |  |
|  | M4 read retains snapshot but not its stamp | U2, U10 | A8, U10, U2 |  |
|  | M5 absent cache gets built by write through | U1, A8 | A8, U1, U13 |  |
|  | M6 rebuild class change rebuilt inside writer | U5, A6 | A6, U5 |  |
|  | M7 noop still persists | U4 | U4, U6 |  |
|  | M8 write through restamps without reparse | A3, A4, A3ref | A3, A3b, A3ref, A4, A7, A7b, A8, U10, U11, U12, U2, U3, U6, U8, U9 |  |
|  | M11 write through asks for the merged triples (second full copy) | U12 | U12 |  |
|  | M12 retained inferred layer deserialized regardless | U12 | U12 |  |
|  | M13 explicit triples built eagerly for the return value | U12 | U12 |  |
|  | M9 delta entry written with fresh stamp but no triples (the #3788 class) | A3, A4 | A10, A2, A2b, A3, A3b, A3ref, A4, A5, A7, A7b, A8, U10, U11, U12, U2, U3, U6, U8 |  |
|  | M10 write through is a noop | A3, A4, A8, U2, U3 | A2b, A3, A3b, A3c, A3ref, A4, A6, A7, A7b, A8, U10, U11, U12, U2, U3, U4, U5, U6, U7, U8, U9 |  |
| `packages/cli/src/cache/loadVaultTriples.ts` (control red: 0) | L5 caller adapter ignored on full parse | L1 | L1 | 0 |
|  | L6 vault mismatch guard removed | L3 | L3 |  |
|  | L1 write through failure propagates to the command | A5 | A5, L4 |  |
|  | L2 caller owned cache manager ignored | A8 | A8, L2, L3 |  |
|  | L3 hit notice mislabelled as delta | A3, A4, A9 | A10, A3, A4, A8, A9, L5 |  |
|  | L4 loader result drops the cache manager | A3, A4 | A2b, A3, A3b, A3c, A3ref, A4, A5, A6, A9, L2 |  |
| `packages/cli/src/commands/apply.ts` (control red: 0) | P5 write through only after success (L1 regression) | A3c | A3c, A9 | 0 |
|  | P6 dry run counts as executed | A9 | A9 |  |
|  | P1 flag ignored full parse | A3, A4, A9 | A10, A2, A2b, A3, A3b, A3c, A3ref, A4, A5, A6, A9 |  |
|  | P2 write through never runs | A3, A4, A3b, A3c | A2b, A3, A3b, A3c, A3ref, A4, A5, A6, A9 |  |
|  | P7 write through flag ignored (always writes through) | A10 | A10 |  |
|  | P8 write through without use cache not refused | A11 | A11 |  |
|  | P4 load notice on stdout | A2, A9 | A10, A2, A3, A3b, A3ref, A4, A5, A9 |  |
| `packages/cli/src/commands/resolve-buttons.ts` (control red: 0) | R1 flag ignored full parse | A9 | A2b, A9 | 0 |
|  | R2 flag always on | A1 | A1, A2b, A9 |  |
|  | R3 notice into the json stdout | A2, A9 | A2, A2b, A9 |  |
| `packages/cli/src/commands/create.ts` (control red: 0) | C2 validator not told about the cache | A2, A8, A9 | A2, A8, A9 | 0 |
|  | C3 write through removed from create | A8 | A8, A9 |  |
|  | C4 write through flag ignored by create (always writes through) | A9 | A9 |  |
|  | C5 create write through without use cache not refused | A11c | A11c |  |
| `packages/cli/src/services/CandidateShaclValidator.ts` (control red: 0) | V1 use cache ignored | A8 | A2, A8, A9 | 0 |
|  | V2 cache manager not forwarded | A8 | A8 |  |
| `CacheManager.ts under the multi-process harness` (control red: 0) | PM1 write through is a noop (next process pays the delta) | P2, P3, P4 | P2, P3, P4, P7, P8 | 0 |
|  | PM2 restamp without reparse (stale precondition, the #3788 class) | P3 | P3, P4, P5, P7, P9 |  |
| `apply.ts under the multi-process harness` (control red: 0) | PP1 flag ignored full parse | P2, P3, P4 | P2, P3, P4, P7, P9 | 0 |
|  | PP2 write through never runs | P2, P3, P4 | P2, P3, P4, P7 |  |
|  | PP3 write through flag ignored (always writes through) | P9 | P9 |  |
|  | PP4 write through without use cache not refused | P10 | P10 |  |
| | | | | |
| `cache-manifest-delta-4263.spec.json` — the #4263 spec, whole, on this branch (shared subject `CacheManager.ts`; control red: 0) | 32 mutants, every one reddens its `expect`; M12 (write in place, not atomic — re-anchored in this PR) → A13 | — | 32/32 distinguish | 0 |

Reading it: `expected` is the axis the mutant was written to catch; `driver RED` is everything that actually went red (a superset is the normal case — a mutant that kills the write-through also kills every axis downstream of it; what the driver requires is expected ⊆ red and control = 0). Guards with no mutant: the `create` lazy-load guard (`if (useCache)` around the CacheManager import — under the `--write-through` gate a constructed-but-unused CacheManager has no observable effect; documented in `create.ts` and `create.spec.json` `_note_C1`) and the `return` after `process.exit` in the two refusals (a test-harness artefact: the jest `process.exit` mock records instead of exiting).

Axes: **A1–A11c** (`tests/integration/use-cache-apply-write-through-4264.integration.test.ts`, 18 — real commands via `parseAsync` on a temp vault, real converter + adapter + CacheManager; "separate process" = a fresh command instance per step sharing only `triples.json`), **U1–U12** (`tests/unit/cache/CacheManager.writeThrough.test.ts`, fake converter, call-count axes), **L1–L5** (`tests/unit/cache/loadVaultTriples.test.ts`), **P1–P10** (`tests/integration/use-cache-4264-proc-chain.harness.ts` — genuine `spawn` of the built `dist/index.js`: index → 3 apply processes with `--write-through` → twin no-flag chain byte-compare → resolve-buttons → create → delta-only chain → refusal; local + under the driver, not in CI since the CLI job has no built dist). All counts are call counts / file bytes / exit codes — never wall-clock.

## Known limits

- Documented divergence: on a cache that `index` built (`inferenceEnabled: true`) the flagged store additionally carries the inferred layer (RDFS `Instance_class` closure + prototype-chain inheritance), exactly as `query --use-cache` does — a precondition or binding on an inherited / inferred triple can evaluate differently from the no-flag full parse (axis A2b, req AC2 + non-goal). `index --no-inference` gives the explicit graph; the write-through keeps whichever layer the cache has.
- `index` peak RSS main vs this branch was not compared (only the `--use-cache` readers in the N2 table were); `index` on the branch measured 1.22 GB / 2.43 GB (no inference / inference) in the matrix, no main figure taken.
- `create --use-cache` without `--validate` under `--write-through` reads the whole cache to fold in one file (L3) → #4267. Not on the default path.
- A same-size rewrite within the filesystem's mtime resolution is invisible to the manifest (delta and write-through alike) — not the case on APFS / ext4; named in the req, not solved here.
- The remaining concurrency window is a writer renaming between this process's `stat` and its own `rename` — the lock-free window #4263 already accepts; self-corrected by the next reader's manifest diff / the next `index`.
- Consumer wiring (`poll_assistant.py` / `kitelev-exo`: `--use-cache`, dropping the pre-turn `index --force`) is a follow-up outside this repo; `--use-cache` / `--write-through` stay opt-in.
- `create`'s default-path cost (frontmatter / shape scans) is not a triple-cache problem — separate item.

## Checks

- CLI jest, full package (`packages/cli`, 180 suites): 179 green; the one red suite is `sparql-exo003.integration.test.ts › mixed format support` and it fails identically against the **published 16.240.19** bundle (swapped `dist/index.js`, same 1/10) — pre-existing on main, not touched here. The #4264 suites + `create.test.ts` + the #4263 integration suite: 120/120.
- `node scripts/check-cli-types.mjs` / `node scripts/check-test-types.mjs`: no new type errors (baselines 12 / 426 pairs).
- `npm test` locally: the obsidian-plugin batch stops on `VaultSettingsRegistry.test.ts › registry ↔ graph parity`, which needs the `packages/exoas-exo` data submodule that this worktree does not have checked out (CI checks submodules out recursively) — environmental, not a code result; core / plugin code is untouched by this PR. `test:e2e:docker` needs Docker, which is not available on this machine — e2e runs as the CI shards.
- `@req:cb707868-356f-495d-825a-182e66ba8bcd` bindings are full UUIDs on every axis; `requirements-trace` stays green (the req is Approved until merge, then flipped to Active).

https://claude.ai/code/session_019RhXyeZjCvDZizn6qZcWeF


kitelev:
Post-release smoke — published `@kitelev/exocortex-cli@16.241.0` on a fresh `cp -a` copy of `vault-bot-kitelev` (`smoke-4264.py`, never the live vault): `index --force`, then the bot chain twice — `--use-cache` (delta-only default) and `--use-cache --write-through` — plus the bare `--write-through` refusal.

```
✅ S0 published CLI 16.241.0
✅ S1 index --force rc=0
✅ S2 create --use-cache rc=0 load=[triple cache: hit] wt=[None]
✅ S3 move --use-cache rc=0 load=[triple cache: delta (1 file(s) re-parsed)]
✅ S4 start --use-cache rc=0 load=[triple cache: delta (1 file(s) re-parsed)]
✅ S5 create --write-through rc=0 load=[triple cache: delta (1 file(s) re-parsed)] wt=[write-through persisted (1 file(s) re-parsed)]
✅ S6 move --write-through rc=0 load=[triple cache: hit] wt=[write-through persisted (1 file(s) re-parsed)]
✅ S7 start --write-through rc=0 load=[triple cache: hit]
✅ S8 resolve-buttons --use-cache rc=0 load=[triple cache: hit]
✅ S9 bare --write-through rc=2
PASS=10 FAIL=0
```

Requirement `cb707868-356f-495d-825a-182e66ba8bcd` flipped to **Active** (kitelev/exoas-exo-reqs@380674a, `implementedBy: PR kitelev/exocortex#4270 (squash f557e7f9)`; Verification wording for C4/A9 per review r3 L-b).

PR #4301, opened 2026-09-20
fix(core,cli): a POSIX bracket expression in a body is not a wikilink (#4219)

### LINKED ISSUE #4219: cli: POSIX bracket expressions ([[:space:]]) in asset body parsed as wikilinks — create emits junk exo__Asset_bodyLink (14 in vault-exodev), set-body rejects the same body
## Наблюдаемое

Тело ассета содержит POSIX bracket expression из bash-примера (`grep -qE '(^|[[:space:]])--help'`). Два пути записи тела ведут себя противоположно:

| путь | исход |
|---|---|
| `exocortex-cli create --body-file …` | принимает; при индексации `[[:space:]]` парсится как wikilink ⇒ триплет `exo__Asset_bodyLink` с целью `:space:` |
| `exocortex-cli set-body <тот же файл> --body-file …` (любая правка тела, даже не трогающая этот фрагмент) | отказывает: `❌ Error: Wikilink [[:space:]] — file not found in vault` |

Замер 2026-09-09, vault-exodev (`query --no-cache`):

```sparql
PREFIX exo: <https://exocortex.my/ontology/exo#>
SELECT ?s ?l WHERE { ?s exo:Asset_bodyLink ?l .
  FILTER(CONTAINS(STR(?l), ':space:') || CONTAINS(STR(?l), ':alpha:') || CONTAINS(STR(?l), ':digit:') || CONTAINS(STR(?l), ':alnum:')) }
```

→ **14 мусорных bodyLink** (канарейка: всего `bodyLink` 8001). Носители — тела правил корпуса, вынесенные в граф пилотом «правила из графа» (`exoas-aiknow`: 794f1844, c824325e, dfe75afc, f0585b6d, …): их текст цитирует bash-паттерны с POSIX-классами по построению.

## Почему это дефект CLI, а не авторинга

`[[:name:]]` — POSIX character class, а не wikilink: цель вида `:слово:` невозможна ни в filename-, ни в uid-схеме. Токенизатор wikilink обязан её пропускать в обеих точках (индексация `bodyLink` и валидация `set-body`), иначе:

- `create` тихо портит граф (мусорные рёбра, SHACL их не судит);
- `set-body` делает тело неправимым без `--skip-wikilink-validation`, а флаг снимает проверку со ВСЕГО тела (корпусный floor `vault-asset-creation` §седьмой подвид запрещает его как решение).

## Предложение (код)

1. В общем wikilink-токенизаторе (индексация + валидатор) исключать совпадения `\[\[:[a-z]+:\]\]`.
2. Тест: тело с `[[:space:]]`/`[[:alpha:]]` ⇒ `bodyLink` не эмитится, `set-body` принимает; тело с настоящим `[[<uid>]]` ⇒ поведение прежнее (revert-verify).
3. Пере-индексация уберёт 14 существующих рёбер без миграции данных.

https://claude.ai/code/session_01AKDHoFopwtnkmV4bwmLFou

kitelev:
Closes #4219.

## Symptom — the same text failed in two OPPOSITE directions

`[[:space:]]` inside a quoted bash pattern is `[` followed by the character class `[:space:]`, but the wikilink tokenizer read it as a link to `:space:`:

| path | outcome |
|---|---|
| indexing | emitted `exo__Asset_bodyLink → ":space:"` — a junk edge no SHACL shape judges, so it accumulated **silently** |
| `set-body` | **REFUSED** the same body (`Wikilink [[:space:]] — file not found`, rc=3), so a note that merely quotes a grep pattern became uneditable without `--skip-wikilink-validation` — a flag that drops validation for the **whole** body |

Reproduced on published **16.241.14** in an isolated vault before touching any code:

```
A. indexing  → Found 2 result(s):  ":space:"  ":alpha:"
B. set-body  → rc=3  ❌ Wikilink [[:space:]] — file not found in vault
```

The defect is **active, not historical**: the issue counted 14 junk edges on 2026-09-09; the same query on 2026-09-20 returns **17** out of 8627 body links (canary non-empty, so the zero-risk is measured rather than assumed).

## The predicate is a SHAPE, not a list of class names

Grouping the actual junk in the graph gave exactly two forms:

| value | count |
|---|---|
| `:space:` | 16 |
| `:слово:` | 1 |

The second is a **documentation example inside the corpus itself** — non-Latin, and an allow-list of the twelve POSIX class names would have missed it while being exactly as wrong a link. Hence `^:[^\s[\]|]+:$`.

That this cannot drop a **working** edge was measured, not argued:

```
assets whose exo__Asset_label has the shape :x:  → 0
files with a colon in the name                    → 0
```

A target wrapped in colons is unreachable in either naming scheme, so skipping it removes junk only.

## One predicate, both sides

`isPosixBracketExpression` lives in core and is called by **both** the indexer (`NoteToRDFConverter.extractBodyWikilinks`) and the CLI's `WikilinkValidator`. Two private copies would be free to drift into disagreeing about what counts as a link — and *"the indexer and the validator agree"* is the very property this fix establishes. What is not indexed is not validated either, by construction.

## Revert-verify — each prediction written down BEFORE the run

| mutation | predicted | actual |
|---|---|---|
| guard removed in the **indexer** | 1 red | core **1 failed / 309 passed** ✅ |
| guard removed in the **validator** | 3 red / 2 green | cli **2 failed / 3 passed** ❌ |

⛤ **The mismatch was the finding.** My axis *"a resolving link next to `[[:digit:]]`"* mocked `findFileByLinkpath` to resolve **every** argument, so `:digit:` "resolved" too and the axis stayed green with the guard gone — **vacuous**. The mock now resolves only the real target; re-run gives the predicted **3 failed / 2 passed**. Had I trusted the first number, a dead axis would have shipped looking like coverage.

## No regression

- `packages/core`: **397 suites / 8997 passed**, 0 failed (as CI runs it)
- `packages/cli`: **2210 passed**, 0 failed
- `npm run check:types`: **0 errors**
- the three CI `lint` guards: green · `spec-anchor-sweep`: 0 broken anchors

Two local-environment notes, both resolved by execution rather than assumption:
- three `dynamic-commands` suites failed until I ran `git submodule update --init` (they read vault fixtures from `packages/exoas-exocmd`; CI checks out `submodules: recursive`) — green afterwards, 120/120;
- `sparql-exo003.integration.test.ts` spawns the built `dist/index.js` and is `describe.skip`ped under `CI=true` (verified: 10 skipped). Its one residual local failure is the known legacy-no-uid case, and **cannot** be caused by this diff — its fixture contains **zero** `[[:x:]]` targets, which is the only thing the change skips.

## Review note

Touches `packages/core/src/services/NoteToRDFConverter.ts` — a parser path — so per the repo's auto-merge discipline this is **not** `--auto` merged: code-reviewer before a manual squash.


kitelev:
## Review applied — MEDIUM fixed in `e4fc0545`

The MEDIUM is real and I reproduced it independently before fixing (real regex + predicate, both call sites side by side):

| body | indexer skips | validator skips |
|---|---|---|
| `[[:space:]]` | true | true |
| `[[ :space: ]]` | **FALSE** | true |
| `[[ Note A ]]` | false | false |

The validator trims the captured target before asking; the indexer asked about the **raw** capture, and the predicate is anchored (`^:`). So on a padded form the verdict split — and in the worse direction of both: validation got laxer while the graph kept being polluted. That is exactly the defect #4219 exists to close, left in place for the padded variant, and unguarded by any axis. Good catch.

**Fix:** the DECISION now uses `match[1].trim()`.

⛔ The **value** added to the index stays the raw capture. Trimming it too would silently change how every padded link (`[[ Note A ]]`) is stored — a wider behaviour change than this PR is scoped to make, and one no axis asked for. A second axis locks it: `[[ Note A ]]` still indexes as `" Note A "`.

**Three axes added** — padded skip on the indexer side, padded accept on the validator side, and the untrimmed-value ratchet.

**Revert-verify:** dropping the `.trim()` from the decision → core **1 failed / 311 passed**, the failure being exactly the new padded axis. Restored → 312/312.

**After the fix:** core **397 suites / 8999 passed** · cli **2211 passed** · `check:types` clean.

**LOW accepted as-is.** The "cannot lose a working edge" claim is corpus-measured, and the docstring already says so ("Measured on vault-exodev") rather than asserting universality. The reviewer independently traced every resolver path — uid, basename, label, alias, path-form, date note — and found no structural form that produces the `:x:` shape; only a deliberately `:tag:`-shaped label or alias could, which is a recognizable code idiom rather than a naming anyone reaches for. Narrowing the predicate further would cost the non-Latin case that is already in the corpus.

Issue #3115, opened 2026-05-16
fix(cli): SHACL converter emits exo__Asset_prototype wikilink as xsd:string literal instead of IRI


kitelev:
## Summary

SHACL-lite pre-validator rejects valid wikilink values for `exo__Asset_prototype`, classifying them as `xsd:string` literals instead of IRIs. Blocks creation of any new `ems__Task` instance from a prototype — i.e. the entire recurring-task instantiation flow (used by `/week-planner`, `/smoke-log`, `/oshia-log`, weekly БАД pillbox-fill, etc.).

## Repro (deterministic, ~10 sec)

```bash
TMP=/tmp/test-prototype-task.md
cat > "$TMP" <<'INNER'
---
exo__Asset_isDefinedBy: "[[!kitelev]]"
exo__Asset_uid: 5655ef7b-1f67-4dbc-91e5-2e82ce82f4b9
exo__Asset_createdAt: 2026-05-16T15:07:06+0500
exo__Asset_updatedAt: 2026-05-16T15:07:06+0500
exo__Instance_class:
  - "[[ems__Task]]"
ems__Effort_status: "[[ems__EffortStatusBacklog]]"
exo__Asset_prototype: "[[f2dccb6a-802d-48d3-8e8a-2c4264197692]]"
exo__Asset_label: Test
exo__Asset_relates:
  - "[[499fc4df-d5cb-4d2e-a724-05e2d6cd2ed5|2026-W21]]"
---
INNER
cat "$TMP" | ~/.claude/hooks/lib/shacl-pre-validate.sh \
  "/Users/kitelev/vault-2025/03 Knowledge/kitelev/5655ef7b-1f67-4dbc-91e5-2e82ce82f4b9.md"
# exit=1
# - [shape?] sh:datatype violation: literal "f2dccb6a-802d-48d3-8e8a-2c4264197692"
#   has datatype <http://www.w3.org/2001/XMLSchema#string>,
#   expected https://exocortex.my/ontology/exo#Asset
```

## Expected behavior

Per project-wide wikilink invariant (settled 2026-05-14):

> Любая конструкция `[[...]]` в vault — это **ссылка на ассет**. Никогда не литерал. Без контекстных исключений. `[[uuid|label]]` → IRI = uuid. `[[uuid]]` → IRI = uuid.

Converter should emit:
`<this-asset> exo:Asset_prototype <https://exocortex.my/ontology/asset/f2dccb6a-...>`

[...]

Issue #2102, opened 2026-02-10
feat(rdf): Dual storage for UUID-based wikilinks (File IRI + UUID Literal)


kitelev:
## Summary
Enable SPARQL queries to find assets by UUID literals by storing both File IRI and UUID Literal for UUID-based wikilinks in RDF conversion.

## User Story / JTBD
**As a** SPARQL query author
**I want** to search for assets using UUID literals like `?s exo:Asset_prototype "e3347bcf-bb50-4fb7-9064-14266469384b"`
**So that** I can reliably find all instances of a prototype without relying solely on file path resolution

## Background
**Implementation Plan Reference:** RFC `[[2f37bf21-fc8b-4343-8abd-22cebe33b6b9|RFC: Dual Storage for UUID-based Wikilinks]]`

Currently, `NoteToRDFConverter.valueToRDFObject()` processes UUID-based wikilinks like:
```yaml
exo__Asset_prototype: "[[e3347bcf-bb50-4fb7-9064-14266469384b]]"
```

And stores ONLY the File IRI:
```turtle
<...> exo:Asset_prototype <obsidian://vault/.../e3347bcf-bb50-4fb7-9064-14266469384b.md> .
```

This causes SPARQL queries searching for UUID literals to fail. Real-world impact:
- `find + grep` found 103 breakfast instances
- SPARQL query returned only ~5 instances

## Related Issues
- Part of Milestone: N/A (standalone improvement)
- blockedBy: None
- Enables: More reliable SPARQL queries for asset relationships

## BoK References
| Body of Knowledge | Chapter/Section | Relevance |
|-------------------|-----------------|-----------|
| SWEBOK v3 | Ch. 2 Software Design | RDF data model design for dual representation |
| SWEBOK v3 | Ch. 3 Software Construction | Refactoring valueToRDFObject() method |
| SWEBOK v3 | Ch. 4 Software Testing | Unit tests for both IRI and Literal storage |
| DMBOK v2 | Ch. 5 Data Modeling | Semantic representation of asset relationships |

### Architecture Context
**Affected Layer:** Infrastructure (RDF Conversion)
**File:** `packages/exocortex/src/services/NoteToRDFConverter.ts`
**Method:** `valueToRDFObject()` at lines 467-526

### Implementation Steps
1. Add `isUUID(value: string): boolean` helper method to detect UUID patterns
2. Modify `valueToRDFObject()` return type to `Promise<(IRI | Literal)[]>`
[...]

PR #4304, opened 2026-09-20
fix(core): a datatype range entry does not constrain IRI nodes (#4268)

### LINKED ISSUE #4268: bug(shacl-lite): 16.240.18 flags string-ranged prefix__Name tokens as sh:class violations (0→9 on vault-exodev, exoas-flow status models) — regression or latent data defect surfaced by #4266
## Summary
`validate schema --shapes-mode` on vault-exodev: **16.240.16 → 0 violations / 545 warnings**, **16.240.18 → 9 violations in 2 nodes / 546 warnings** — same vault, nothing changed on disk between the two runs (pinned-version A/B, 2026-09-18 13:20 +05). The delta lands with #4266 (CURIE-literal range `xsd:<local>` now reaches `sh:datatype` via `loadFromRDFGraph`).

Both flagged nodes are `exoas-flow` status-model assets whose **string-ranged** properties hold TBox-form tokens (`prefix__Name`, no wikilink):

```
assetspaces/kitelev/exoas-flow/flow/ee81b677-bf07-4666-8a92-2d6676a03753.md  (flow__StatusModelEffort)
  flow__StatusModel_statusProperty: ems__Effort_status
  flow__StatusModel_activeStatus:   ems__EffortStatusDoing
  flow__StatusModel_boardStatus:    [ems__EffortStatusBacklog, ems__EffortStatusDoing, ems__EffortStatusDraft]
assetspaces/kitelev/exoas-flow/flow/3840e11a-745a-4960-8606-b4c382da6375.md  (RFC status model: exodev__RFC_status / RFCStatusDraft|Proposed|Approved)
```

Reported as e.g. `[Violation] sh:class violation: <https://exocortex.my/ontology/ems#EffortStatusDoing> does not conform to expected class http://www.w3.org/2001/XMLSchema#string`.

Mechanism: `NoteToRDFConverter.valueToRDFObject` expands a bare `prefix__Name` string into a **symbolic IRI** (`isClassReference → expandClassValue`, ~`:1657-1662`), so the object is an IRI even though the author wrote a string; the property's declared range is `xsd:string`; after #4266 the shape enforces `sh:datatype xsd:string` and the IRI object fails. Before #4266 the datatype constraint never materialised, so the mismatch was latent.

## Why it matters
Vault CIs run `validate schema` with `@latest` → they go red at the next CLI bump without any data change. Two readings are possible and the issue is to pick one:
1. **Validator over-reach** — string-ranged properties legitimately hold `prefix__Name` tokens (the flow ontology uses them as *names* of statuses/properties, not links); `sh:datatype` should not be enforced against converter-synthesised IRIs, or `xsd:string` range should be tolerated for symbolic-expanded values.
2. **Data/TBox defect** — the properties should be object properties with a class range (or the values written as `"[[uid]]"`), and the flow ontology needs a fix + migration.

## Acceptance Criteria (BDD)
- [ ] **Given** vault-exodev at the state of 2026-09-18 **When** `validate schema --shapes-mode` runs on the fixed CLI (or fixed data) **Then** 0 violations, and the resolution (1 or 2) is named in the PR body.
- [ ] **Given** a string-ranged property with a bare `prefix__Name` value **When** validated **Then** the outcome is the same as on 16.240.16 (reading 1) OR the asset is migrated and the shape stays strict (reading 2) — one of them, documented.
- [ ] Revert-verify: the 2-node repro above is a test fixture; the fix flips it RED→GREEN.

## Technical Notes
- Repro: `npx -y @kitelev/exocortex-cli@16.240.16 validate schema --shapes-mode --vault ~/vaults/vault-exodev` vs `@16.240.18`.
- Affected: `packages/core/src/…/shacl` (loadFromRDFGraph, #4266), `NoteToRDFConverter.valueToRDFObject`, `exoas-flow` ontology (`flow__StatusModel_*` ranges).
- Related: #4266 (introduced the datatype enforcement), #4263/#4265 (same release; not the cause — A/B isolates to validation).

## Size
- [x] M (≤4h)

kitelev:
Closes #4268.

## Which of the two readings — and why the mechanism decides it, not taste

The issue offers two: **(1) validator over-reach** or **(2) data/TBox defect**. This PR ships **(1)**, and the deciding fact is not a preference:

> The IRI branch checked the value for membership in `http://www.w3.org/2001/XMLSchema#string`. **Nothing is an instance of a datatype**, so that check cannot go green for *any* data — it is false by construction, not strict.

A check with no reachable green outcome is a defect of the checker regardless of what the data says. The file itself already states the rule it was breaking, at `XSD_DATATYPE_PREFIXES`:

> *"Range entries with these prefixes are sh:datatype constraints (apply to Literals). **All other range entries are sh:class constraints (apply to IRI nodes only)**."*

The Literal branch implements its half — it filters the range down to datatype entries and skips silently when none remain, with the reasoning spelled out (`#2102`, `#3115`). The IRI branch never got the mirror. This PR adds it.

It also blames the wrong actor: the node kind there is **not authored but inferred**. `valueToRDFObject → isClassReference` expands a bare `prefix__Name` string into a symbolic IRI *without consulting the declared range*, so an IRI under a datatype-only range says nothing about what the author wrote.

## Reading (2) is a real but separate defect — measured, not waved off

I enumerated every property definition in the three canonical vaults (deduped by `exo__Asset_uid`):

| metaclass × range form | count |
|---|---|
| ObjectProperty × wikilink (class) | 366 |
| DatatypeProperty × CURIE `xsd:*` | 136 |
| other coherent | 9 |
| **ObjectProperty × CURIE `xsd:*`** | **4** |
| | **515 total** |

The two coherent combinations are 502 of 515 (97.5 %). The four outliers are `flow__StatusModel_{statusProperty,activeStatus,boardStatus}` — the three that produce all 9 violations — plus `ems__AlcoholicDrinkSession_drink` (latent: its values are ordinary strings, so they emit as literals). So the TBox **does** lie, and that is worth fixing; but it is an ontology decision about the `flow` status models (their values reference *two different* status classes, `ems#EffortStatus` and `exodev#RFCStatus`), not a reason to keep a tautologically-false check in the engine. Filed separately rather than smuggled in here.

## Measured outcome

| vault | 16.240.16 | 16.240.18 (reported) | this build |
|---|---|---|---|
| exodev | 0 viol / 545 warn | **9 viol in 2 nodes** / 546 | **0 / 545** |
| my | — | 2 viol (`ems#Effort`, unrelated) | **2** — preserved |
| tbank | — | 0 | 0 |

vault-my is the control that matters: those two are *genuine* class violations against a real class range, and they survive. The fix removes a check that could not pass, not checking.

## Revert-verify — prediction written down BEFORE the run, 5/5 matched

`packages/core/tests/services/shacl-datatype-range-4268.spec.json`, run via `mutant-driver.py`. Guards enumerated from the diff per `integration-test-revert-verify` §A8: the filter expression, the skip guard, and the three sites that report the surviving expectation.

| mutant | predicted red | actual |
|---|---|---|
| M1 filter removed (pre-fix behaviour) | T71 T72 T73 | ✅ same |
| M3 skip guard removed | T71 | ✅ |
| M4 violation-message join | T72 | ✅ |
| M5 unresolvable-ref message join | T73 | ✅ |
| M6 `expectedRange` join | T72 T73 | ✅ |
| control (unmutated) | — | 0 red |

**T74 is the control for the untouched Literal branch and stays green under every mutant** — the stronger claim. Its non-vacuity is not assumed: it *actually went red* during development, when its first fixture self-tagged the literal (`'not-a-number'^^xsd:integer` under range `xsd:integer`) and `literalConformsToDatatype` short-circuits to `true` on `literalDatatype === expected`, never reaching the lexical judgement. The committed fixture tags the literal `xsd:string` so that path runs.

One conceivable mutation carries no mutant, and the reason is collateral rather than equivalence — inverting the filter polarity strips the *class* entries instead, emptying every class-only range in the suite; its red set would be dominated by collateral and it locks nothing M1 does not. Documented in the spec's `_doc`.

## No regression

- `packages/core` as CI runs it (`--testPathIgnorePatterns '/tests/performance/'`, ci.yml:891): **399 suites / 9019 passed**, rc=0.
- Without that exclusion, 3 wall-clock perf assertions fail locally (`TripleStorePerformance`, `MultiHopPrototypeChainPerformance`) — the category CI excludes by design; unrelated to this diff.
- `npm run check:types`: rc=0.
- The three CI `lint` guards (`check-test-antipatterns.sh`, `validate-shard-assignments.mjs`, `check-coverage-monotonic.mjs`): all rc=0.
- `spec-anchor-sweep.py` over the touched subject: 1 spec on it (mine), 0 broken anchors — no foreign spec anchors on `ShaclLiteValidator.ts`.
- eslint on the two changed files reports 3 warnings at lines 5/6/49 — byte-identical on `origin/main`, i.e. pre-existing; my block adds none.

## Review note

Touches `packages/core/src/services/ShaclLiteValidator.ts` — a schema/validation path — so per the repo's auto-merge discipline this is **not** `--auto` merged: code-reviewer before a manual squash.

https://claude.ai/code/session_01VDLpni8LrBAbuWew4fnMPK


kitelev:
## On AC3 — why the repro is a unit + live-vault axis and NOT an integration fixture

The issue asks for "the 2-node repro above is a test fixture". I built one and **measured that it would be vacuous**, so I am not shipping it; here is the measurement rather than an opinion.

A minimal synthetic vault reproducing the exact shape — `probe__Model_activeStatus` typed `exo__ObjectProperty` with `exo__Property_range: xsd:string`, an instance whose value is the bare token `probe__StatusDoing`, and that token's individual typed by a class — run on the **defective published CLI** (`@latest`, which still has the bug):

```
✅ Loaded 55 triples
✅ Vault conforms to all shapes (no violations).
ℹ️  1 warning(s) (1 unresolvable class/symbolic ref(s) — not validated)
```

A **warning**, not a violation — so that fixture is green with the fix and green without it.

⛤ And I did not take that at face value either: "no violation" is also what you get when the shape never loaded. Canary — give the same Single-cardinality property two values:

```
⚠️  Found 1 SHACL violation(s) in 1 node(s):
      [Violation] sh:maxCount violation: expected at most 1 value for <…/probe#Model_activeStatus>, got 2
```

The shape **is** loaded and the property **is** under it. The range check simply takes the unresolvable-ref path, because in a small synthetic vault the symbolic IRI `probe#StatusDoing` is not a typed subject — whereas in vault-exodev `ems#EffortStatusDoing` is, which is precisely why the live corpus got `sh:Violation` and the fixture cannot.

So the regression is pinned where it is actually observable:
- **unit** — T71–T73 with the value explicitly typed, so the unresolvable-ref path (green either way) cannot swallow the axis; 5 mutants, prediction written before the run, 5/5 matched;
- **live corpus** — vault-exodev 9 → 0 with warnings back to 545 (the 16.240.16 number), and vault-my keeping both of its genuine `ems#Effort` violations as the control that the fix does not over-filter.

Building a fixture whose symbolic value is a resolvable subject may be possible, but I could not do it without replicating the term-IRI indexing that makes it so — and a fixture that gets that subtly wrong is exactly a vacuous axis wearing the costume of coverage.

kitelev:
## The blast radius is wider than CI noise — this could gate ExoSync merges

While checking downstream consumers of `constraint` / `expectedRange` I found one that does **not** treat these as cosmetic. `packages/core/src/services/sync/MergeShaclGate.ts:122`:

```ts
if (v.constraint === "class") {
  // … class-range results must reach that decision regardless of the severity
  if (v.actualValue !== undefined && !isMounted(v.actualValue)) {
    skippedOpenWorld++;   // open-world: unmounted ref ≠ violation
    continue;
  }
  violations.push(v);     // ← gates the merge
  continue;
}
if (v.severity !== "sh:Violation") continue;  // non-class warnings don't gate
```

Class-constraint results **bypass the severity filter by design** (#3488) and gate the merge whenever the value is *mounted*. The false violations this PR removes were exactly that shape: `constraint: 'class'`, `actualValue` = `ems#EffortStatusDoing`, and that value **is** a typed subject in vault-exodev — which is precisely why it surfaced as `sh:Violation` rather than the unresolvable-ref warning.

⇒ on the defective build, any asset carrying a `flow__StatusModel_*` predicate would reach the ExoSync merge gate with a class-constraint entry that passes `isMounted` and lands in `violations`. I am stating the code path, not a sighting: I have not observed an actual blocked merge, and I did not try to induce one on live sync state.

Nothing is lost in the other direction — the entries this PR stops producing were, for a datatype-only range, judgements that could not have been green anyway.

Other consumers of the narrowed field are pass-through only: `CandidateShaclValidator.ts:207` copies `expectedRange`, `ShaclConformanceError.ts:63` prints it. The plugin's `PropertyValidationService` has an unrelated local `expectedRange` from a SPARQL result — not this field. No consumer parses it semantically, so narrowing it to the class half is safe.

kitelev:
## Review round 1 — APPROVE, 0 CRITICAL/HIGH/MEDIUM, 2 LOW. Both applied in `62ace8b7`.

### LOW-1: my justification was an overclaim, and the reviewer falsified it by probe — not by reading

I wrote, in the code comment **and** the commit message **and** the PR body above: *"that check can never go green for any data."*

That is false as stated. The reviewer built a fixture where the value node's own `exo__Instance_class` points at a datatype IRI, so the pre-fix comparison `vc === expectedClass` evaluates **true** — and ran it against the reverted code: **0 violations**, i.e. green. The absolute claim is falsifiable.

⛔ **was:** "…is false BY CONSTRUCTION — nothing is an instance of `xsd:string` — so that check can never go green for any data."
✅ **mechanism:** the check asks whether the value is an instance of `xsd:string`, which **no well-formed vault asset can be** — so for well-formed data it had no reachable green outcome. The falsifying input is corrupt (a subject typed as a datatype), and **both revisions treat it identically**, which is why the conclusion survives while the justification did not.

The correction is written into the comment next to the claim, not quietly softened — a durable artifact that has already said something false should show what it said, or the next reader re-derives it. It also stood in this PR body and the first commit message, which is why it is restated here rather than only in the diff.

### LOW-2: an accepted loss of signal, now recorded at the guard

Before the fix, an IRI under a pure-datatype range that was **also** unresolvable (a genuinely dangling reference, no type triple at all) produced the `sh:class unresolvable-ref` warning. After the fix, `classRanges.length === 0` short-circuits first, so that case now yields **no signal at all** — the same treatment as a perfectly resolvable value.

This is consistent with the fix (the class check does not apply to a datatype-only range, whatever the value's resolvability), and the reviewer found no consumer or gate depending on that warning. But it is a real if minor loss, so it is now stated at the guard itself, with the dangling-reference signal for these predicates assigned to the TBox fix in #4305 rather than left implicit.

### What the reviewer verified independently rather than taking from me

- Re-ran the suite on the exact committed state: **142/142**.
- Re-ran `mutant-driver.py` on the shipped spec: **all 5 mutants matched their declared `expect` sets exactly**, control 0 red.
- Reproduced the historical T74 bug (self-tagging the literal `xsd:integer`) to check my "non-vacuous control" claim — **T74 goes red**, so the control does exercise `literalConformsToDatatype`'s lexical path.
- Probed the `isExternalOntologyIRI` ordering concern I raised in the brief: harmless and **pre-existing**, unaffected by this diff (a pure class range with an external value skips for the same, unmodified reason).
- Traced `ShapeLoader`'s two range-extraction paths to confirm both emit datatype IRIs in a form `XSD_DATATYPE_PREFIXES` recognises — no form mismatch between what the loader can produce and what the filter detects.
- Re-derived the `MergeShaclGate` consequence independently and reached the same reading as my earlier comment: the fix can only make that gate **less** likely to over-block.

`62ace8b7` is comment-only: core 399 suites / 9019 passed, typecheck 0, `spec-anchor-sweep` 0 broken anchors.

